Issue #17, opened 2022-03-09
How about making the server delete expired files by itself?


svenstaro:
Currently you need to run an outside command to make sure expired files are actually deleted. Couldn't this step be easily performed by the server itself during runtime so you don't need to add the complexity of having a second thing that modifies the files?

orhun:
Hey!

Let me explain how expiration works first.

Currently, expiry dates are tracked by the file extensions which are basically [Unix Epoch](https://en.wikipedia.org/wiki/Unix_time). For example:

```sh
curl -F "file=@file.txt" -H "expire:10s" paste.example.com
```

`file.txt` will appear on the server side as e.g. `file.txt.1647033013063` so running [this command](https://github.com/orhun/rustypaste#cleaning-up-expired-files) is usually enough for removing this type of files.

About your proposal, I think it makes sense to have such functionality built-in. This will also make it possible to actually check the expiration status of files instead of blindly running `rm *.[0-9]*`.

I see a couple of ways of doing it:

1. Start a separate thread along with the server thread and clean these files up based on an interval that is set via configuration file.
2. Option 1 but instead of using intervals, integrate a file retention period based on file size. (this is what [0x0](https://github.com/mia-0/0x0) does)
3. When an expired file is visited, remove it immediately from the server side. This is fairly simple to implement than 1 and 2.

What do you think?

svenstaro:
I think option 1. makes the most sense.

shumvgolove:
In case someone find this useful, I wrote this little script:
```sh
#!/bin/env sh
now=$(date +%s)
find /path/to/rusty/ -maxdepth 2 -type f -iname "*.[0-9]*" |
while read filename; do
	[ "$(( ${filename##*.} / 1000 - ${now} ))" -lt 0 ] && rm ${filename}
done
```

Replace `/path/to/rusty` with your folder and put this script in `cron`.

orhun:
Thanks for sharing the script @shumvgolove! Would you be potentially interested in submitting a PR to update README.md since this looks more applicable than my one-liner?

I'm eventually going to implement auto-deletion of expired files but I think this is useful for people who would like to disable that feature.

shumvgolove:
> Would you be potentially interested in submitting a PR to update README.md

Sure thing!

orhun:
[...]

PR #20, opened 2022-03-16
docs(readme): add shell script for cleaning files

shumvgolove:
As suggested in https://github.com/orhun/rustypaste/issues/17#issuecomment-1068510377


orhun (approved):
Thank you!

PR #23, opened 2022-06-24
chore(config): do not check for duplicate files by default

plantroon:
It is an expensive operation to do on slower hardware and can take unreasonable amount of time for bigger files


orhun (approved):
Seems reasonable.

Issue #13, opened 2022-02-07
Landing page


orhun:
Show a cool 😎 landing page instead of redirecting to the repository when `/` is visited.

Some ideas:

- [http://ix.io/](http://ix.io/) (man page style)
- [https://0x0.st/](https://0x0.st/) (something simple)

TheTechRobo:
Should it be configurable or included as a const?

orhun:
> Should it be configurable or included as a const?

It should be both I think, we can have a default landing page and have it configurable in the config file. But handling it as a const is a good start.

PR #26, opened 2022-10-02
feat(server): add landing page

TheTechRobo:
Fixes orhun/rustypaste#13

The default landing page is set as a const in the binary, but is configurable with the server.landing_page key. Should the key be in another section (i.e. not `[server]`? Handling is added for when the config file doesn't have any landing page. I updated the default config.toml.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -7,6 +7,18 @@ address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
 timeout="30s"
+landing_page = """
+Submit files via HTTP POST here:
+    curl -F 'file=@example.txt' <server>"
+This will return the finished URL.
+
+The server administrator might remove any pastes that they do not personally
+want to host.
+
+If you are the server administrator and want to change this page, just go
+into your config file and change it!
+a
+"""
 
 [paste]
 random_url = { enabled = true, type = "petname", words = 2, separator = "-" }
--- a/config.toml
+++ b/config.toml
@@ -7,8 +7,7 @@ address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
 timeout="30s"
-landing_page = """
-Submit files via HTTP POST here:
+landing_page="""Submit files via HTTP POST here:
     curl -F 'file=@example.txt' <server>"
 This will return the finished URL.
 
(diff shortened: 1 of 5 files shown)


TheTechRobo:
I kinda want to add a `{}` you can do to insert the repository URL. But I'm not sure how to format that - Rust doesn't accept `format!(a_string, another_string)` where `a_string` is `Hello, {}!` and `another_string` is `world`

TheTechRobo:
I just did that, lmk if I shouldn't.

TheTechRobo:
Maybe I should also add how soon pastes expire.

orhun:
> The default landing page is set as a const in the binary, but is configurable with the server.landing_page key. Should the key be in another section (i.e. not `[server]`? Handling is added for when the config file doesn't have any landing page. I updated the default config.toml.

I think it is fine 👌🏼 

> I kinda want to add a `{}` you can do to insert the repository URL. But I'm not sure how to format that - Rust doesn't accept `format!(a_string, another_string)` where `a_string` is `Hello, {}!` and `another_string` is `world`
> 
> Should I just do a `replace` on the string?

Hmm, it looks fine for now. I might change the behavior after this PR is merged though. I'm not sure if I want duplicate landing page text as both constant and in the config file. However, that's a concern for later.

> Maybe I should also add how soon pastes expire.

That would be nice, please add that as well 👍🏼

orhun:
CI is failing, can you update tests and apply formatting?

TheTechRobo:
> I'm not sure if I want duplicate landing page text as both constant and in the config file. However, that's a concern for later.

I thought about that, but then won't existing configurations break?

> > Maybe I should also add how soon pastes expire.
> That would be nice, please add that as well 👍🏼

Right now it's hardcoded. I'm not sure what the best way to go about this is.

TheTechRobo:
> CI is failing, can you update tests and apply formatting?

All sorted.

TheTechRobo:
It appears the only test I broke was test_index, which makes sense. Every other failing test is broken on master.

orhun:
> It appears the only test I broke was test_index, which makes sense (since that's all I changed). Every other failing test is broken on master.

Tests should be run with `cargo test --test-threads 1`. `test_index` was failing because the config was not given via `app_data` while initializing the server. I fixed it in ee93fbc761a62a0ed98b4a12aee3e524ebb22a2f

I also changed the behavior in c91ffecbaa00ddb2f2c374169f37905868e1d253, I think it makes sense to redirect to the GitHub repository (as before) if landing page is not specified.

orhun (approved):

TheTechRobo:
> test_index was failing because the config was not given via app_data while initializing the server.

I knew that one, just wasn't sure how to fix it. Thanks for the --test-threads tip and thanks for merging!

PR #37, opened 2023-05-17
feat(deps): allow using openssl for tls

nekopsykose:
this allows reusing the openssl present on a distro already, and takes a release build with it from 5.5 to 4.7mb

### DIFF
--- a/.github/workflows/shuttle.yml
+++ b/.github/workflows/shuttle.yml
@@ -24,7 +24,7 @@ jobs:
         run: cargo binstall -y cargo-shuttle
       - name: Prepare for deployment
         shell: bash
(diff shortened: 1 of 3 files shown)


orhun:
Thanks for your contribution!

orhun (approved):
bors r+

inglor:
Is this a choice which can be made during build? Can this be documented?

orhun:
@inglor documented in 8359aadec68cfcf88fcc91f214438eeaa8765ca8

nekopsykose:
i think you need --no-default-features to actually use it, but yes

orhun:
ah, right :D fixed

PR #48, opened 2023-05-27
feat(server): allow configuring the content type for landing page

### LINKED ISSUE #39: Add web UI
It would be super handy for when I’m on mobile to have a basic web interface where I could give the application a URL or file and get back a link.

magikid:
I realized that I don't need to implement this in a repo at all.  The user can already set a custom landing page in the config.  The only thing I needed to change was the content type so that the browser rendered it.  This makes the landing page content type available in the configuration.

Fixes #39


orhun (approved):
Thanks for your contribution!

PR #51, opened 2023-05-30
docs(example): add information about using HTML form

magikid:
This adds some information and example config file for creating an HTML form in the server.


orhun:
Thank you for the PR! This made me think, maybe we can add a `landing_page_file` for specifying a more extended HTML file for the landing page. I think it wouldn't be convenient to put 200 lines of HTML in the config file, what do you think?

Created an issue about this here: #52

orhun (approved):
Neato! Did some improvements in 36917a6f3fe2661a9b068e7114d692e460a8e6a5

PR #61, opened 2023-06-12
feat(server): honor X-Forward-* headers

### LINKED ISSUE #54: Support for Forwarded Headers
I just came across this awesome project and was able to bring it up in no time.

During testing, I discovered that there is currently no feature to support Forwarded request headers, e.g.

- `X-Forwarded-For`
- `X-Forwarded-Host`
- `X-Forwarded-Proto`

This would be really useful to have for setups where the service is running behind a reverse-proxy or gateway and the possibility to adjust the logging output based on their availablity, to have the real IP addresses of the clients available in the log.

Thank you for creating such a useful tool and keep up the great work. :heart:

tessus:
Behind a reverse proxy, the log entries always showed the IP address of the reverse proxy. With this change the real IP address of the client is shown. Since the IP address is only used for info in the log, there are no security implications.

closes #54

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -157,7 +157,7 @@ async fn main() -> IoResult<()> {
         App::new()
             .app_data(Data::clone(&config))
             .app_data(Data::new(http_client))
-            .wrap(Logger::default())
+            .wrap(Logger::new("%{r}a \"%r\" %s %b \"%{Referer}i\" \"%{User-Agent}i\" %T"))
             .wrap(ContentLengthLimiter::new(
                 server_config.max_content_length.get_bytes(),
             ))
@@ -197,7 +197,7 @@ async fn actix_web(
             web::scope("")
                 .app_data(Data::clone(&config))
                 .app_data(Data::new(http_client))
-                .wrap(Logger::default())
+                .wrap(Logger::new("%{r}a \"%r\" %s %b \"%{Referer}i\" \"%{User-Agent}i\" %T"))
                 .wrap(ContentLengthLimiter::new(
                     server_config.max_content_length.get_bytes(),
                 ))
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -157,7 +157,7 @@ async fn main() -> IoResult<()> {
         App::new()
             .app_data(Data::clone(&config))
             .app_data(Data::new(http_client))
-            .wrap(Logger::default())
+            .wrap(Logger::new("%{r}a \"%r\" %s %b \"%{Referer}i\" \"%{User-Agent}i\" %T"))
```
orhun, line 160: Can you share an example of how this log entry looks like?

orhun: Also, what was the reason behind that change?

tessus: If my IP address is `1.2.3.4` and I am using a reverse proxy (as I mentioned in https://github.com/orhun/rustypaste/issues/54#issuecomment-1586487208), the log file will show the following:

```
[2023-06-12T02:22:15Z INFO  actix_web::middleware::logger] 127.0.0.1 "GET / HTTP/1.1" 200 271 "-" "Mozilla/5.0 Gecko/20100101 Firefox/114.0" 0.000036
```

With this PR, the log will show:

```
[2023-06-12T02:22:15Z INFO  actix_web::middleware::logger] 1.2.3.4 "GET / HTTP/1.1" 200 271 "-" "Mozilla/5.0 Gecko/20100101 Firefox/114.0" 0.000036`
```

orhun: Got it. Just checked the [default logger format](https://docs.rs/actix-web/latest/actix_web/middleware/struct.Logger.html#method.default) and it makes sense!

orhun (approved):
Thank you for your contribution!

tessus:
You are welcome. I love rustypaste. Thanks for writing it!

PR #63, opened 2023-06-12
docs(readme): list all supported units for expiry

tessus:
These suffixes are also supported in the `config.toml`.

---

The duration is not consistently used. e.g. in the `config.toml` it says `default_expiry = "1h"` and not `default_expiry = "1hours"` as it is shown in the README. Also, I think people are rather used to single character duration units like s, m, h, d, w, y.

So I believe listing all supported units is useful for usability and consistency.


orhun (approved):
Fair enough 👍🏼

Issue #52, opened 2023-05-30
Landing page improvements and refactoring


orhun:
A landing page example was added in #51 and it demonstrates how to use a HTML form for uploads. To extend this, we can introduce a new option called `landing_page_file` which will read the contents of a file to serve it as the landing page. The reasoning behind this is that an user might not want to put X lines of HTML code into the configuration file.

```toml
[server]
# landing_page = ""
landing_page_file = "index.html"
landing_page_content_type = "text/html; charset=utf-8"
```

Another idea is to refactor the landing page configuration into a separate section:

```toml
[landing_page]
# text = ""
file = "index.html"
content_type = "text/html; charset=utf-8"
```

I think this would be cleaner but we can go with the first option as the first stage of the implementation. More comments needed.

vihu:
Just a suggestion, it would also be nice to allow a user of the landing page to provide server side auth_token in the input form.

orhun:
This is now implemented in #65 

> Just a suggestion, it would also be nice to allow a user of the landing page to provide server side auth_token in the input form.

Feel free to submit a new issue for tracking!

PR #64, opened 2023-06-13
feat(server): support a file for the landing page

tessus:
As contemplated in #52, this PR adds another config option `landing_page_file`.

If the file cannot be found, it will fall back to `landing_page`.

Tested: manually with 
- relative path
- absolute path
- file not found
- neither `landing_page_file` or `landing_page` set

Issue #52 also talks aboutrefactoring and changing the example html landing_page, thus I am not sure whether this PR should close that issue.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -33,10 +33,10 @@ async fn index(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
     match &config.server.landing_page_file {
         Some(file) => {
             let file_contents = fs::read_to_string(file).unwrap_or("".to_string());
-            if file_contents.is_empty() == false {
+            if !file_contents.is_empty() {
                 return Ok(HttpResponse::Ok()
                     .content_type(content_type)
-                    .body(file_contents.clone()));
+                    .body(file_contents));
             }
         },
         None => (),
--- a/src/server.rs
+++ b/src/server.rs
@@ -29,6 +29,19 @@ async fn index(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
         .landing_page_content_type
         .clone()
         .unwrap_or("text/plain; charset=utf-8".to_string());
+
+    match &config.server.landing_page_file {
+        Some(file) => {
+            let file_contents = fs::read_to_string(file).unwrap_or("".to_string());
+            if file_contents.is_empty() == false {
+                return Ok(HttpResponse::Ok()
+                    .content_type(content_type)
+                    .body(file_contents.clone()));
+            }
+        },
+        None => (),
+    }
+
     match &config.server.landing_page {
         Some(page) => Ok(HttpResponse::Ok()
             .content_type(content_type)
--- a/src/server.rs
+++ b/src/server.rs
@@ -38,7 +38,7 @@ async fn index(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
                     .content_type(content_type)
                     .body(file_contents));
             }
-        },
+        }
         None => (),
     }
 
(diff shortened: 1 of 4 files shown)


src/server.rs
```diff
@@ -29,6 +29,19 @@ async fn index(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
         .landing_page_content_type
         .clone()
         .unwrap_or("text/plain; charset=utf-8".to_string());
+
+    match &config.server.landing_page_file {
+        Some(file) => {
```
orhun, line 34: This can simply be replaced with a `if let` statement.

src/server.rs
```diff
@@ -29,6 +29,19 @@ async fn index(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
...
+            let file_contents = fs::read_to_string(file).unwrap_or("".to_string());
+            if !file_contents.is_empty() {
```
orhun, line 36: Let's handle the error gracefully via `?` operator here.

orhun (changes requested):
Can you add unit tests for this?

src/server.rs
tessus: I am rather new to Rust. No idea how to use a `if let` statement instead. Some of he constructs are still alien to me.
I'll have to read up on that.

src/server.rs
tessus: Same here. No idea how to use the `?` operator.

I'm more than happy to get help. It took me already hours to figure out how to read a file, so that the program does not panic when the file is not found. In C this is all very easy. In Rust it makes my head hurt.

src/server.rs
orhun: No worries, it can be refactored as:

```rs
    if let Some(file) = &config.server.landing_page_file {
        let file_contents = fs::read_to_string(file).unwrap_or("".to_string());
        if !file_contents.is_empty() {
            return Ok(HttpResponse::Ok()
                .content_type(content_type)
                .body(file_contents));
        }
    }

```

So that you will get rid of `None => ()` 😃

tessus: Ah, I see. Thanks.

src/server.rs
tessus: I will read up on the lang reference. Thanks.

tessus:
Ah, I see. Nice refactoring.

While I was writing the test cases it showed that it doesn't fall back to the contents of `landing_page` anymore, when the file is not found. Was this intended?

tessus:
I can also start working on this idea of yours:

```toml
[landing_page]
# text = ""
file = "index.html"
content_type = "text/html; charset=utf-8"
```

tessus:
@orhun please advise. If this PR is ok, I can either create a separate one for a new config section or add it to this one. What do you think?

src/server.rs
```diff
@@ -368,6 +374,50 @@ mod tests {
...
+                .app_data(Data::new(RwLock::new(config)))
+                .service(index),
+        )
+        .await;
+        let request = TestRequest::default()
+            .insert_header(("content-type", "text/plain"))
+            .to_request();
+        let response = test::call_service(&app, request).await;
+        assert_eq!(StatusCode::FOUND, response.status());
+        //assert_body(response.into_body(), "landing page").await?;
```
orhun, line 417: Can you remove this comment?

orhun:
Looks pretty good!

>  If this PR is ok, I can either create a separate one for a new config section or add it to this one. What do you think?

A new PR would be more appropriate after we merge this :)

src/server.rs
tessus: done.

I only kept it in case you wanted the fallback to `landing_page`....

orhun (approved):
LGTM!

PR #65, opened 2023-06-16
refactor(config): use a separate section for the landing page

tessus:
Migration path:

Old:

```
[server]
landing_page = "Landing page text."
landing_page_file = "index.html"
landing_page_content_type = "text/html; charset=utf-8"
```

New:

```
[landing_page]
text = "Landing page text."
file = "index.html"
content_type = "text/html; charset=utf-8"
```

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -9,7 +9,9 @@ max_content_length = "10MB"
 upload_path = "./upload"
 timeout = "30s"
 expose_version = false
-landing_page = """
+
+[landing_page]
+text = """
 ┬─┐┬ ┬┌─┐┌┬┐┬ ┬┌─┐┌─┐┌─┐┌┬┐┌─┐
 ├┬┘│ │└─┐ │ └┬┘├─┘├─┤└─┐ │ ├┤
 ┴└─└─┘└─┘ ┴  ┴ ┴  ┴ ┴└─┘ ┴ └─┘
@@ -31,7 +33,8 @@ changed this.
 Check out the GitHub repository at https://github.com/orhun/rustypaste
 Command line tool is available  at https://github.com/orhun/rustypaste-cli
 """
-landing_page_content_type = "text/plain; charset=utf-8"
+#file = "index.txt"
+content_type = "text/plain; charset=utf-8"
 
 [paste]
 random_url = { enabled = true, type = "petname", words = 2, separator = "-" }
--- a/fixtures/test-duplicate-file-upload/config.toml
+++ b/fixtures/test-duplicate-file-upload/config.toml
@@ -3,8 +3,6 @@ address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
 
-[landing_page]
-
 [paste]
 random_url = { enabled = true, type = "petname", words = 2, separator = "-" }
 default_extension = "txt"
--- a/fixtures/test-duplicate-file-upload/config.toml
+++ b/fixtures/test-duplicate-file-upload/config.toml
@@ -3,6 +3,8 @@ address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
 
+[landing_page]
+
 [paste]
 random_url = { enabled = true, type = "petname", words = 2, separator = "-" }
 default_extension = "txt"
--- a/fixtures/test-server-landing-page-file/config.toml
+++ b/fixtures/test-server-landing-page-file/config.toml
@@ -2,9 +2,11 @@
 address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
-landing_page="awesome_landing"
-landing_page_file="page.txt"
-landing_page_content_type = "text/plain; charset=utf-8"
+
+[landing_page]
+text="awesome_landing"
+file="page.txt"
+content_type = "text/plain; charset=utf-8"
 
 [paste]
 random_url = { enabled = false, type = "petname", words = 2, separator = "-" }
--- a/src/config.rs
+++ b/src/config.rs
@@ -47,12 +47,12 @@ pub struct ServerConfig {
     pub auth_token: Option<String>,
     /// Expose version.
     pub expose_version: Option<bool>,
-    // START - code to not break current configs. can be removed in the future.
     /// Landing page text.
+    #[deprecated(note = "use the [landing_page] table")]
     pub landing_page: Option<String>,
     /// Landing page content-type.
+    #[deprecated(note = "use the [landing_page] table")]
     pub landing_page_content_type: Option<String>,
-    // END - code to not break current configs. can be removed in the future.
 }
 
 /// Landing page configuration.
--- a/src/config.rs
+++ b/src/config.rs
@@ -15,7 +15,7 @@ pub struct Config {
     pub server: ServerConfig,
     /// Paste configuration.
     pub paste: PasteConfig,
-    /// Landing Psge configuration.
+    /// Landing page configuration.
     pub landing_page: LandingPageConfig,
 }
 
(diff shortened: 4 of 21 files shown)


tessus:
I don't know why the test suite failed.

```
$ cargo test -- --test-threads 1
    Finished test [unoptimized + debuginfo] target(s) in 0.12s
     Running unittests src/lib.rs (target/debug/deps/rustypaste-22392e8954e7e67b)

running 28 tests
test auth::tests::test_check_auth ... ok
test config::tests::test_parse_config ... ok
test file::tests::test_file_checksum ... ok
test header::tests::test_content_disposition ... ok
test header::tests::test_expiry_date ... ok
test mime::tests::test_match_mime_type ... ok
test paste::tests::test_paste_data ... ok
test random::tests::test_generate_url ... ok
test server::tests::test_auth ... ok
test server::tests::test_index ... ok
test server::tests::test_index_with_landing_page ... ok
test server::tests::test_index_with_landing_page_file ... ok
test server::tests::test_index_with_landing_page_file_not_found ... ok
test server::tests::test_payload_limit ... ok
test server::tests::test_upload_duplicate_file ... ok
test server::tests::test_upload_expiring_file ... ok
test server::tests::test_upload_file ... ok
test server::tests::test_upload_oneshot ... ok
test server::tests::test_upload_oneshot_url ... ok
test server::tests::test_upload_remote_file ... ok
test server::tests::test_upload_url ... ok
test server::tests::test_version ... ok
test server::tests::test_version_without_auth ... ok
test server::tests::test_version_without_config ... ok
test util::tests::test_get_expired_files ... ok
test util::tests::test_glob_match ... ok
test util::tests::test_sha256sum ... ok
test util::tests::test_system_time ... ok

test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.68s

     Running unittests src/main.rs (target/debug/deps/rustypaste-e6109393dcb4f21c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests rustypaste

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

fixtures/test-duplicate-file-upload/config.toml
```diff
@@ -3,6 +3,8 @@ address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
 
+[landing_page]
```
orhun, line 6: Is it possible to make this section optional so that we don't have to specify an empty section for disabling the landing page?

src/config.rs
```diff
@@ -15,6 +15,8 @@ pub struct Config {
     pub server: ServerConfig,
     /// Paste configuration.
     pub paste: PasteConfig,
+    /// Landing Psge configuration.
```
orhun, line 18: Typo: `Psge` -> `page`

src/config.rs
```diff
@@ -43,14 +45,22 @@ pub struct ServerConfig {
     pub timeout: Option<Duration>,
     /// Authentication token.
     pub auth_token: Option<String>,
+    /// Expose version.
+    pub expose_version: Option<bool>,
+}
+
+/// Landing Psge configuration.
```
orhun, line 52: Typo: `Psge` -> `page`

src/config.rs
```diff
@@ -43,14 +45,22 @@ pub struct ServerConfig {
...
+#[derive(Debug, Clone, Default, serde::Serialize, serde::Deserialize)]
+pub struct LandingPageConfig {
     /// Landing page text.
-    pub landing_page: Option<String>,
+    //pub landing_page: Option<String>,
+    pub text: Option<String>,
     /// Landing page file.
-    pub landing_page_file: Option<String>,
+    //pub landing_page_file: Option<String>,
```
orhun, line 59: Please remove the unnecessary code comments.

orhun:
> I don't know why the test suite failed.

Nothing related to this PR I reckon.

fixtures/test-duplicate-file-upload/config.toml
tessus: I was thinking about it when adding it to the test case configs, but then I thought that it was rather highly unlikely that someone would not set either `text` or `file` in production.
Thus I saw the error message for not specifying the section as actually helpful.

But yes, I can do that.

tessus: I was to hasty. Setting `pub landing_page: Option<LandingPageConfig>,` creates a whole bunch of errors. It seems I have to deal with nested `Option` types and I have no idea how to handle that.
I tried using a similar approach as what you did in main.rs with config.settings, but I couldn't figure it out.

I'll try again tomorrow. But I am also open to any help.

tessus: I think I managed to make it work, although I am pretty sure there is a better way to do that. 

Additionally I don't know how to fix the tests in the `server.rs`:

```
error[E0609]: no field `text` on type `std::option::Option<LandingPageConfig>`
   --> src/server.rs:365:29
    |
365 |         config.landing_page.text = Some(String::from("landing page"));
    |                             ^^^^

error[E0609]: no field `file` on type `std::option::Option<LandingPageConfig>`
   --> src/server.rs:387:29
    |
387 |         config.landing_page.file = Some(filename.to_string());
    |                             ^^^^

error[E0609]: no field `text` on type `std::option::Option<LandingPageConfig>`
   --> src/server.rs:408:29
    |
408 |         config.landing_page.text = Some(String::from("landing page"));
    |                             ^^^^

error[E0609]: no field `file` on type `std::option::Option<LandingPageConfig>`
   --> src/server.rs:409:29
    |
409 |         config.landing_page.file = Some(filename.to_string());
    |                             ^^^^
```

config.toml
```diff
@@ -31,7 +33,8 @@ changed this.
 Check out the GitHub repository at https://github.com/orhun/rustypaste
 Command line tool is available  at https://github.com/orhun/rustypaste-cli
 """
-landing_page_content_type = "text/plain; charset=utf-8"
+#file = "index.txt"
```
orhun, line 36: Can you rename this to "index.html"?

fixtures/test-server-landing-page-file/config.toml
```diff
@@ -2,9 +2,11 @@
 address="127.0.0.1:8000"
 max_content_length="10MB"
 upload_path="./upload"
-landing_page="awesome_landing"
-landing_page_file="page.txt"
-landing_page_content_type = "text/plain; charset=utf-8"
+
+[landing_page]
+text="awesome_landing"
+file="page.txt"
```
orhun, line 8: HTML pls 🐻

fixtures/test-duplicate-file-upload/config.toml
orhun: Fixed the test failures, thank you for the implementation!

orhun:
Do you think there is a backward compatible way to support the old-style config? It would be really nice if we can somehow avoid a [major bump](https://semver.org/) due to breaking changes.

orhun:
Oh, can you also update the [HTML form example](https://github.com/orhun/rustypaste/blob/master/examples/html_form.toml) according to these changes?

tessus:
> Do you think there is a backward compatible way to support the old-style config? It would be really nice if we can somehow avoid a [major bump](https://semver.org/) due to breaking changes.

Well, we could leave the previous config options and define which options have the priority if both are set. I reckon this is the only way to not introduce a breaking change.

Breaking changes are always annoying and one can use workarounds to delay a hard cut. But at one point you want to remove the "old" code, otherwise it's just an additional maintenance burden. It's up to you. Let me know how you want to handle this.

tessus (commented):
Thanks for the fix. I couldn't figure out how to resolve this!

fixtures/test-duplicate-file-upload/config.toml
tessus: Thanks for the fix. I couldn't figure out how to resolve this!

tessus:
I have now added code so that "old" configs do not break. I did not add the `landing_page_file`, because this was never released.

I added comments what to delete in the future. Should be straight forward.

Please let me know what you think.

orhun:
I added deprecation notices to the respective config fields and also added warning logs for making the migration more apparent for the users. I think that's more appropriate than having code comments.

orhun (approved):

tessus:
The deprecation warnings are great, but unless I am mistaken they are behind this guard:

```
if let Some(mut landing_page) = config.landing_page.clone() {
```

So they will never show up if someone has not set the section `[landing_page]`. This will also break old configs.

orhun:
Ah, thanks for bringing this to my attention! Should be fixed in 40a87c586be49b629b9cf3ac6790a19ec94b0481

PR #69, opened 2023-06-20
feat(server): add random suffix mode

### LINKED ISSUE #41: Random suffix mode
It would  be convenient to have an "intermediate" random name mode: append a random suffix to the filename before the extension. It would both preserve the original filename and prevent collisions. For example, in random alphanumeric string suffix mode Rustypaste could turn `foo.tar.gz` into `foo.eu7f92x1.tar.gz`. It could do the same with pet names.

tessus:
Unfortunately the current implementation does not retain multiple extensions. So if you upload `foo.tar.gz`, the resulting file will be `eu7f92x1.gz`.

This PR includes 2 changes:

- multiple extensions are retained (e.g. `eu7f92x1.tar.gz`)
- added an optional option `suffix_mode` (bool) to `random_url` (result: `foo.eu7f92x1.tar.gz`)

closes #41

### DIFF
--- a/src/paste.rs
+++ b/src/paste.rs
@@ -125,27 +125,28 @@ impl Paste {
 
         // set_file_name does not accept names that contain a . unless it is the first character (e.g. "file.name" not ok, ".file" ok)
         // as a workaround we use set_extension, because it can set multiple extensions (e.g. "tar.gz")
+        // never accept a filename "." - similar to above where an empty string is replaced by file, and - by stdin
         if input == "." {
             input = "file".to_string();
         }
-
+        // Split the string into an array
         let mut v: Vec<&str> = input.split('.').collect();
-
         let mut filename;
         let mut dotfile = false;
-        let first = v[0];
-        if !first.is_empty() {
+        if !v[0].is_empty() {
             filename = v[0].to_string();
         } else {
+            // If the first array element is empty, it means the file started with a dot (e.g.: .foo)
             filename = format!(".{}", v[1]);
+            // Index shifts one to the right in the array for the rest of the string (the extension)
             dotfile = true;
         }
-
         let mut extension;
-        let l = v.len();
-        if l > 1 {
+        if v.len() > 1 {
+            // To get the rest (the extension), we have to remove the first element of the array, which is the filename
             v.remove(0);
             if dotfile {
+                // If the filename starts with a dot, we have to remove another element, because the first element was empty
                 v.remove(0);
             }
             extension = v.join(".");
--- a/src/paste.rs
+++ b/src/paste.rs
@@ -158,8 +158,8 @@ impl Paste {
         }
 
         if let Some(random_text) = config.paste.random_url.generate() {
-            if let Some(suffix_mode) = config.paste.random_url.suffix_mode {
-                if suffix_mode {
+            if let Some(random_suffix) = config.paste.random_url.random_suffix {
+                if random_suffix {
                     extension = format!("{}.{}", random_text, extension);
                 } else {
                     filename = random_text;
@@ -312,7 +312,7 @@ mod tests {
             enabled: true,
             length: Some(4),
             type_: RandomURLType::Alphanumeric,
-            suffix_mode: Some(true),
+            random_suffix: Some(true),
             ..RandomURLConfig::default()
         };
         let paste = Paste {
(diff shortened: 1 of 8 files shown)


orhun:
Can you please add tests to cover these changes?

tessus:
I have to look into it. I did of course my own testing: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fbd05ecc4a7979366febb168fafd43b5

src/paste.rs
```diff
@@ -116,28 +116,61 @@ impl Paste {
...
+            filename = format!(".{}", v[1]);
+            dotfile = true;
+        }
+
+        let mut extension;
+        let l = v.len();
+        if l > 1 {
+            v.remove(0);
+            if dotfile {
+                v.remove(0);
```
orhun, line 149: I have trouble understanding this code. Can you maybe avoid single line declarations (if possible) and remove the unnecessary line breaks (and maybe use comments)? Frankly, ChatGPT might really help refactor this.

tessus: I have never used ChatGPT and I am not willing to use it. I can explain the code though:

```rust
    // This is just a safeguard, in case a filename is sent that is just a dot.
    // It should never happen, but who knows. Paranoia took over, which is why I added this.
    if input == "." {
        input = "file".to_string();
    }

    // Split the string into an array
    let mut v: Vec<&str> = input.split('.').collect();

    let mut filename;
    let mut dotfile = false;
    // This the first element of the array 
    let first = v[0];
    if !first.is_empty() {
        filename = v[0].to_string();
    } else {
        // If the first array element is empty, it means the file started with a .
        filename = format!(".{}", v[1]);
        // Therefore the index shifts one to the right in the array for the rest of the string
        dotfile = true;
    }

    let mut extension;
    let l = v.len();
    if l > 1 {
        // To get the rest (the extension), we have to remove the first element of the array, which is the filename
        v.remove(0);
        if dotfile {
            // as mentioned above, if the file starts with a ., we have to remove another element because the filename is the second element in the array
            v.remove(0);
```

I am sorry that I didn't add any comments. I thought that by reading the code it was clear what the purpose was. Let me know what you think of the comments.
If they are ok, I'll add them and fix the line breaks and the unnecessary `let first = v[0];`

tessus: I think I can improve the comments. I'll do that later today.

tessus: I think that's done.

tessus:
Please let me know, whether I missed anything. Otherwise I'd hope for a 0.11.0 release soon. Haha. ;-)

orhun:
I cleaned up the implementation a little bit for better readability and added fixture tests. Can you also confirm that it works as expected?

tessus:
I'll test within the hour. What I could see from the code at a first glance, it should be the same behavior.

Although I chose `suffix_mode` for a reason. It is an option to `random_url` (and is only relevant when random_url is enabled), so naming the option `random_` is superfluous. e.g. you don't call the other option `random_type` either. The term `suffix` alone is too less specific IMO, but would still be better than `random_suffix`.
But it's your choice of course.

I'll comment on the result of my tests in a bit.

orhun:
Yeah, it makes sense to use `suffix_mode` in that context. Feel free to revert it!

tessus:
Yep, the code works the same, but I found a bug in my code (I missed it in my earlier test cases), which of course propagated to the refactored one:

```
.foo           -> https://rptest.local/paste/.foo.wtNk.
.foo.bar       -> https://rptest.local/paste/.foo.SL2N.bar
.foo.bar.txt   -> https://rptest.local/paste/.foo.vTGL.bar.txt
foo            -> https://rptest.local/paste/foo.0OPe.txt
foo.bar        -> https://rptest.local/paste/foo.g7yj.bar
foo.bar.txt    -> https://rptest.local/paste/foo.kEaw.bar.txt
```

For the first file `.foo`, the resulting URL should be: `https://rptest.local/paste/.foo.wtNk.txt`

I can look into this later today. For now I just reverted the option name.

tessus:
All fixed. Good to go, ;-)

Not sure why the test suite fails. If you run `cargo test` and the fixtures locally, averythig is fine. I guess gh is still recovering.

orhun:
Looks good, thank you!

> I'd hope for a 0.11.0 release soon.

I will work on it tomorrow 🐻

orhun (approved):
🥳🥳🥳

PR #72, opened 2023-06-27
style(server): add new line character to 404 message

oflifurkan:
I noticed a little missing new line character after I ran command for testing how rustypaste works:
`curl http://localhost:8000/sweeping-tahr`
After that, server side returns:
`file is not found or expired :(` without new line character and terminal output looks bad.


orhun (approved):
LGTM!

Issue #14, opened 2022-02-23
How to setup authentication?


Nbelles:
Hello,

I'm really liking the very minimalist-style setup of rustypaste so far. It's so much faster than everything else I've tried lately. 

I am curious if we could get some more documentation about how to use the authentication portion of rustypaste. How do we setup authentication on the server side? I think the documentation on how to use the authentication on the client side is clear but we'll find out once I figure out how to setup the server side authentication.

Thanks!

inglor:
Check https://github.com/orhun/rustypaste/blob/master/.env file

Nbelles:
That's not super helpful because I don't know what to set AUTH_TOKEN to.

inglor:
It can be something random like the output of `openssl rand -base64 32`. Just make sure it's secret and you set the same on the client side.

Nbelles:
Oh so we have to generate our own key that is shared between client and server? All the server does is check the key that the client sent against the one it has locally?

Nbelles:
Is there a way to set the AUTH_TOKEN in the `config.toml` file? I'm currently running mine in a docker container and it would be easier to set it in the config file.

inglor:
docker supports the .env file as per the README.md:
>  --env-file "$(pwd)/.env"

or docker-compose https://github.com/orhun/rustypaste/blob/master/docker-compose.yml#L10

Nbelles:
This might be a stretch but would you know how to setup this in docker on Unraid? I really don't have much experience with custom configurations for docker containers on Unraid. I don't run any commands to start the docker container so I don't have a place that I would be able to add that argument to the start path. I also don't use docker-compose so I don't have a .yml file that I would edit to make that change.

Nbelles:
I think I resolved it [here](https://github.com/SmartPhoneLover/unraid-docker-templates/issues/4#issuecomment-1049348228) actually. Thanks for the help!

orhun:
Hey @Nbelles, I just released `0.6.3` and now you can set the auth token in the config file like this:

```toml
[server]
auth_token=""
```

PR #84, opened 2023-07-06
feat(server): support multiple auth tokens

tessus:
## Description

This change supports the use of multiple authentication tokens.

Example:

```toml
[server]
auth_tokens = [
  "super_secret_token1",
  "super_secret_token2",
]
```

The previously used `AUTH_TOKEN` environment variable can still be used and will be evaluated as well.

## Motivation and Context

Allow to give people access to an otherwise private rustypaste instance.
see https://github.com/orhun/rustypaste/discussions/79

## How Has This Been Tested?

- test cases (cargo test)
- fixtures 
- manual testing on a Linux test VM

## Changelog Entry

````
### Added

- Add support for multiple auth tokens

Example:

```toml
[server]
auth_tokens = [
  "super_secret_token1",
  "super_secret_token2",
]
```

The previously used `AUTH_TOKEN` environment variable can still be used and will be evaluated as well.
````

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/fixtures/test-server-auth-multiple-tokens/test.sh
+++ b/fixtures/test-server-auth-multiple-tokens/test.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-auth_tokens="rustypasteisawesome token1 token2 token4"
+auth_tokens=("rustypasteisawesome token1 token2 token4")
 
 content="topsecret"
 
@@ -12,7 +12,7 @@ run_test() {
   result=$(curl -s -F "file=@file" localhost:8000)
   test "unauthorized" = "$result"
 
-  for auth_token in $auth_tokens
+  for auth_token in ${auth_tokens[@]}
   do
     result=$(curl -s -F "file=@file" -H "Authorization: $auth_token" localhost:8000)
     test "unauthorized" != "$result"
--- a/fixtures/test-server-auth-multiple-tokens/test.sh
+++ b/fixtures/test-server-auth-multiple-tokens/test.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+auth_token="rustypasteisawesome"
+content="topsecret"
+
+setup() {
+  echo "$content" > file
+}
+
+run_test() {
+  result=$(curl -s -F "file=@file" localhost:8000)
+  test "unauthorized" = "$result"
+
+  result=$(curl -s -F "file=@file" -H "Authorization: $auth_token" localhost:8000)
+  test "unauthorized" != "$result"
+  test "$content" = "$(cat upload/file.txt)"
+  test "$content" = "$(curl -s $result)"
+}
+
+teardown() {
+  rm file
+  rm -r upload
+}
--- a/fixtures/test-server-auth-multiple-tokens/test.sh
+++ b/fixtures/test-server-auth-multiple-tokens/test.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-auth_tokens=("rustypasteisawesome token1 token2 token4")
+auth_tokens=("rustypasteisawesome" "token1" "token2" "token4")
 
 content="topsecret"
 
--- a/fixtures/test-server-auth-multiple-tokens/test.sh
+++ b/fixtures/test-server-auth-multiple-tokens/test.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
-auth_token="rustypasteisawesome"
+auth_tokens="rustypasteisawesome token1 token2 token4"
+
 content="topsecret"
 
 setup() {
@@ -11,10 +12,13 @@ run_test() {
   result=$(curl -s -F "file=@file" localhost:8000)
   test "unauthorized" = "$result"
 
-  result=$(curl -s -F "file=@file" -H "Authorization: $auth_token" localhost:8000)
-  test "unauthorized" != "$result"
-  test "$content" = "$(cat upload/file.txt)"
-  test "$content" = "$(curl -s $result)"
+  for auth_token in $auth_tokens
+  do
+    result=$(curl -s -F "file=@file" -H "Authorization: $auth_token" localhost:8000)
+    test "unauthorized" != "$result"
+    test "$content" = "$(cat upload/file.txt)"
+    test "$content" = "$(curl -s $result)"
+  done
 }
 
 teardown() {
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -5,23 +5,12 @@ use actix_web::{error, Error};
 ///
 /// `Authorization: (type) <token>`
 pub fn check(host: &str, headers: &HeaderMap, tokens: Option<Vec<String>>) -> Result<(), Error> {
-    if tokens.is_some() {
-        let mut token_found = false;
+    if let Some(tokens) = tokens {
         let auth_header = headers
             .get(AUTHORIZATION)
             .map(|v| v.to_str().unwrap_or_default())
             .map(|v| v.split_whitespace().last().unwrap_or_default());
-        if let Some(tokens) = tokens {
-            if !token_found {
-                for token in &tokens {
-                    if auth_header.unwrap_or_default() == token {
-                        token_found = true;
-                        break;
-                    }
-                }
-            }
-        }
-        if !token_found {
+        if !tokens.iter().any(|v| v == auth_header.unwrap_or_default()) {
             log::warn!(
                 "authorization failure for {} (header: {})",
                 host,
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -4,23 +4,13 @@ use actix_web::{error, Error};
 /// Checks the authorization header for the specified token.
 ///
 /// `Authorization: (type) <token>`
-pub fn check(
-    host: &str,
-    headers: &HeaderMap,
-    token: Option<String>,
-    tokens: Option<Vec<String>>,
-) -> Result<(), Error> {
-    if token.is_some() || tokens.is_some() {
+pub fn check(host: &str, headers: &HeaderMap, tokens: Option<Vec<String>>) -> Result<(), Error> {
+    if tokens.is_some() {
         let mut token_found = false;
         let auth_header = headers
             .get(AUTHORIZATION)
             .map(|v| v.to_str().unwrap_or_default())
             .map(|v| v.split_whitespace().last().unwrap_or_default());
-        if let Some(token) = token {
-            if !token.is_empty() && auth_header.unwrap_or_default() == token {
-                token_found = true;
-            }
-        }
         if let Some(tokens) = tokens {
             if !token_found {
                 for token in &tokens {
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: fixtures/test-server-auth-multiple-tokens/test.sh [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+
+setup() {
+  echo "$content" > file
+}
+
+run_test() {
+  result=$(curl -s -F "file=@file" localhost:8000)
+  test "unauthorized" = "$result"
+
+  result=$(curl -s -F "file=@file" -H "Authorization: $auth_token" localhost:8000)
```
orhun, line 14: Let's also test access with alternative tokens ("token1", "token2" etc.) here.

src/config.rs
```diff
@@ -45,6 +45,8 @@ pub struct ServerConfig {
     pub timeout: Option<Duration>,
     /// Authentication token.
     pub auth_token: Option<String>,
+    /// Authentication tokens.
+    pub auth_tokens: Option<Vec<String>>,
```
orhun, line 49: How about we deprecate `auth_token` field and only support `auth_tokens`? I think it would help greatly refactoring the code and also having an array is definitely more convenient. What do you think?

fixtures/test-server-auth-multiple-tokens/test.sh
tessus: sure.

src/config.rs
tessus: Hah. Yep, I was thinking about this as well. But then I have noticed something interesting: it seems that this config option was never made public. It's not in the example `config.toml`, nor in the README. IMO it is perfectly fine to deprecate it (and get rid of it when you create a `1.0.0` release in the future), but I still think that the env var should be kept. Especially for the Docker folks who like to configure as much as possible with env vars.

You don't even have to mention the deprecation, because this option was never mentioned in te first place either. Well, this only works, if you keep the env var.

In many cases a single auth token _might_ still be the most prevalent use-case, thus keeping the env var could be a good thing. But it's up to you.

orhun: I think it was once mentioned in the changelog and release page which is enough for visibility. Also, this was a requested feature in #14 

We should keep the environment variable for sure. I meant `server.auth_token` -> `server.auth_tokens` and showing a log message if `server.auth_token` is used in the configuration file.

tessus: Yep, this makes perfect sense. I'll add that.

fixtures/test-server-auth-multiple-tokens/test.sh
tessus: done

src/config.rs
tessus: Done.

tessus:
Hmm, sorry for the formatting error. I am using a pre-commit hook, so I don't know why it didn't trigger. Very strange.... and annoying.

I have also added an improvement. The current state only printed deprecation warnings when an action happened. e.g. when the `/` was requested and the deprecated option was read from the config.

I have kept the same logic for the new deprecation warning, but I also added these warnings at server startup.

tessus:
Argh! (\*pulling my hair out\*)

I ran `cargo clippy`, which apparently does not include the tests.

fixtures/test-server-auth-multiple-tokens/test.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+auth_tokens="rustypasteisawesome token1 token2 token4"
```
orhun, line 3: I think we should use a bash array here:

```bash
auth_tokens=("rustypasteisawesome" "token1" "token2" "token4")
```

src/auth.rs
```diff
@@ -4,22 +4,41 @@ use actix_web::{error, Error};
...
-                    "authorization failure for {} (header: {})",
-                    host,
-                    auth_header.unwrap_or("none"),
-                );
-                return Err(error::ErrorUnauthorized("unauthorized"));
+pub fn check(
+    host: &str,
+    headers: &HeaderMap,
+    token: Option<String>,
+    tokens: Option<Vec<String>>,
```
orhun, line 11: Can we pass a single argument here (only `tokens`) instead?

src/main.rs
```diff
@@ -48,6 +49,17 @@ fn setup(config_folder: &Path) -> IoResult<(Data<RwLock<Config>>, ServerConfig,
...
+        log::warn!("[server].auth_token is deprecated, please use [server].auth_tokens");
+    }
+    if config.server.landing_page.is_some() {
+        log::warn!("[server].landing_page is deprecated, please use [landing_page].text");
+    }
+    if config.server.landing_page_content_type.is_some() {
+        log::warn!(
+            "[server].landing_page_content_type is deprecated, please use [landing_page].content_type"
+        );
+    }
```
orhun, line 62: Maybe extract these into a new function called `warn_deprecation` or something like that. What do you think?

src/auth.rs
tessus: This would be a lot easier after the token was removed from the code (release 1.0.0). Otherwise I have to merge token and tokens before passing them into the check function. I can look into it though. In theory this should be easy, but my lack of Rust knowledge makes this is a challenge:

I don't know how to pass a reference of config into a function.

e.g.

```rust
let config = Config::parse(&config_path).expect("failed to parse config");

let tokens = merge_tokens(&config);
```

What is the type of config? I don't know how to declare the function. 

```
fn merge_tokens(config: &what-to-put-here) -> Option<Vec<String>> 
```

But I will look into it. This gives me the opportunity to learn more about Rust.

src/main.rs
tessus: Sure, but since this ciode will be deleted as soon as the deprecated stuff is removed, I didn't bother. Also I run into the same issue here as described here https://github.com/orhun/rustypaste/pull/84#discussion_r1264681653

But I'll look into it. If you have a quick answer as to what the type of config is, so that I can pass it into a function, I can do this in a few minutes. ;-)

Otherwise it will take a bit to figure this out. I also have a few other things planned for today so I might have to do that tomorrow.

fixtures/test-server-auth-multiple-tokens/test.sh
tessus: Sure, but why? It doesn't give me any advantage. I am just curious.

I love arrays. The reason why I am careful with them is that macOS comes with an ancient version of bash (3.something). I am using a MacPorts bash, but with newer macOS it is even harder to replace `/bin/bash` even with SIP off.
Thus I usually use `#!/usr/bin/env bash`, but sometimes I forget and use `#!/bin/bash` - and then all hell breaks lose. ;-)
Unless I need associative arrays or want to do fancy array stuff, I just use the above used construct.

But I am more than happy to change it. One array coming up.

src/main.rs
tessus: Hmm, I think it might be `Config`. At least the rust-analyzer says so. I'll run a few tests.

tessus: Hmm, that was easier than expected. Done.

fixtures/test-server-auth-multiple-tokens/test.sh
tessus: done

src/auth.rs
tessus: I think I figured it out. I'll work on it.

tessus: Done. It was remarkably easy. I think I am starting to understand Rust a bit better.

fixtures/test-server-auth-multiple-tokens/test.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+auth_tokens=("rustypasteisawesome token1 token2 token4")
```
orhun, line 3:
```suggestion
auth_tokens=("rustypasteisawesome" "token1" "token2" "token4")
```

src/auth.rs
```diff
@@ -4,22 +4,31 @@ use actix_web::{error, Error};
...
+            log::warn!(
+                "authorization failure for {} (header: {})",
+                host,
+                auth_header.unwrap_or("none"),
+            );
+            return Err(error::ErrorUnauthorized("unauthorized"));
+        }
     }
     Ok(())
 }
```
orhun, line 34: I think this method can be simplified as:

```rs
pub fn check(host: &str, headers: &HeaderMap, tokens: Option<Vec<String>>) -> Result<(), Error> {
    if let Some(tokens) = tokens {
        let auth_header = headers
            .get(AUTHORIZATION)
            .map(|v| v.to_str().unwrap_or_default())
            .map(|v| v.split_whitespace().last().unwrap_or_default());
        if !tokens.iter().any(|v| v == auth_header.unwrap_or_default()) {
            log::warn!(
                "authorization failure for {} (header: {})",
                host,
                auth_header.unwrap_or("none"),
            );
            return Err(error::ErrorUnauthorized("unauthorized"));
        }
    }
    Ok(())
}
```

src/server.rs
```diff
@@ -306,6 +297,27 @@ pub fn configure_routes(cfg: &mut web::ServiceConfig) {
...
+            merged.insert(0, token.to_string());
+        }
+        if let Ok(env_token) = env::var(AUTH_TOKEN_ENV) {
+            merged.insert(0, env_token);
+        }
+        Some(merged)
+    } else {
+        None
+    }
+}
```
orhun, line 319: Please move this method to make it a part of `Config`. I think calling `config.get_tokens()` is cleaner.

fixtures/test-server-auth-multiple-tokens/test.sh
tessus: Oops, my bad. Thanks forcatching this!

src/auth.rs
tessus: Nice, let me run some tests. e.g. does the `tokens.iter().any` loop stop, when one is found. e.g. I don't want to loop over a huge array, if a token was matched in the beginning of the array already...
This is why I kept my longer construct. I thought about perf.

src/server.rs
tessus: I'll look into it.

tessus: Done. Btw, I have done the same for the deprecation warning function.

src/auth.rs
tessus: Yep, the loop is discontinued as soon as a token is found. Thus this code change is awesome. Makes the code a lot shorter. Nice. Thanks for the tip.

tessus: Done

src/config.rs
```diff
@@ -107,6 +112,42 @@ impl Config {
...
+            }
+            if let Ok(env_token) = env::var(AUTH_TOKEN_ENV) {
+                merged.insert(0, env_token);
+            }
+            if !merged.is_empty() {
+                return Some(merged);
+            }
+        }
+        None
+    }
```
orhun, line 135: This function can be refactored as:

```rs
    pub fn get_tokens(&self) -> Option<Vec<String>> {
        let mut tokens = self.server.auth_tokens.clone().unwrap_or_default();
        if let Some(token) = &self.server.auth_token {
            log::warn!("[server].auth_token is deprecated, please use [server].auth_tokens");
            tokens.insert(0, token.to_string());
        }
        if let Ok(env_token) = env::var(AUTH_TOKEN_ENV) {
            tokens.insert(0, env_token);
        }
        (!tokens.is_empty()).then_some(tokens)
    }
```

tessus: Wow, I think I still have to get more into the spirit of Rust.

orhun (approved):
Thank you for your contribution!

tessus:
Thank you for your help and guidance. I always learn something new thanks to you.

orhun:
Anytime!

Issue #28, opened 2022-10-03
Mitigating the effects of content discovery and fuzzer tools


orhun:
Even though the names of uploaded files are random, [fuzzer](https://en.wikipedia.org/wiki/Fuzzing) utilities can be used to discover content on the server.

Related questions are:

- How can `rustypaste` be affected by such situation?
  - Need a demonstration of fuzzing with [rustbuster](https://github.com/phra/rustbuster) or a similar tool.
- Should there be any precautions for preventing this?
  - Implement blocking user-agents for common fuzzer tools (?)
- Should we do anything?

I think it is an interesting topic and needs some brainstorming for improving security.

PR #94, opened 2023-07-13
feat(server): add an endpoint for retrieving a list of files

ajbdev:
## Description

Hi there! I forked rustypaste to add a feature I need for a frontend I am building utilizing rustypaste as a backend. It occurred to me that it may be beneficial for the upstream project, although I am not sure if the feature fits with the philosophy or roadmap of the project. If it doesn't feel free to decline this PR and I'll understand completely.

Please note: I am a total Rust beginner and may have implemented this completely wrong. Feel free to critique accordingly.

## Motivation and Context

I'm using rustypaste as a lightweight fileshare backend on my local network and I need a way to list all files for a frontend UI component I am building. This is also a common feature of other pastebin services and I think it may have many potential use cases beyond my own. 

I only listed files in the upload dir and did not list urls, one shot files, or one shot urls. For my use case this is all I need but I could change this if requested. 

## How Has This Been Tested?

I've tested it locally on my machine and written a test that covers functionality.

### Added

- Add a function that lists all files in the upload directory and transforms them into a JSON HTTP response of objects representing the files in the directory.
- Add new variable to config.toml to expose the json list endpoint. By default, it is not enabled.

### Changed
- Modified the behavior of the `GET /{file}` route to conditionally execute and return the transformation function if the JSON index path configuration value matches the name of the requested file.

## Types of Changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other 

## Checklist:

- [?] My code follows the code style of this project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
 serde_json = "1.0.100"
+chrono = "0.4.26"
 
 [dependencies.config]
 version = "0.13.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ actix-files = "0.6.2"
 awc = { version = "3.1.1" }
 env_logger = "0.10.0"
 log = "0.4.19"
-serde = "1.0.166"
+serde = "1.0.167"
 futures-util = "0.3.28"
 petname = { version = "1.1.3", default-features = false, features = [
   "std_rng",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
-serde_json = "1.0.100"
+serde_json = "1.0.103"
 uts2ts = "0.1.0"
 
 [dependencies.config]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,12 +16,7 @@ include = ["src/**/*", "Cargo.*", "LICENSE", "README.md", "CHANGELOG.md"]
 default = ["rustls"]
 openssl = ["actix-web/openssl", "awc/openssl"]
 rustls = ["actix-web/rustls", "awc/rustls"]
-shuttle = [
-  "dep:shuttle-actix-web",
-  "dep:shuttle-runtime",
-  "dep:shuttle-static-folder",
-  "dep:tokio",
-]
+shuttle = ["dep:shuttle-actix-web", "dep:shuttle-runtime", "dep:shuttle-static-folder", "dep:tokio"]
 
 [dependencies]
 actix-web = { version = "4.3.1" }
@@ -52,6 +47,7 @@ shuttle-actix-web = { version = "0.20.0", optional = true }
 shuttle-runtime = { version = "0.20.0", optional = true }
 shuttle-static-folder = { version = "0.20.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
+serde_json = "1.0.100"
 
 [dependencies.config]
 version = "0.13.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
-uts2ts = "0.1.0"
+uts2ts = "0.3.0"
 
 [dependencies.config]
 version = "0.13.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,6 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
-serde_json = "1.0.103"
 uts2ts = "0.1.0"
 
 [dependencies.config]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ shuttle = [
   "dep:shuttle-actix-web",
   "dep:shuttle-runtime",
   "dep:shuttle-static-folder",
-  "dep:tokio"
+  "dep:tokio",
 ]
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
-uts2ts = "0.3.0"
+uts2ts = "0.4.0"
 
 [dependencies.config]
 version = "0.13.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
 serde_json = "1.0.100"
-chrono = "0.4.26"
+uts2ts = "0.1.0"
 
 [dependencies.config]
 version = "0.13.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,9 @@ default = ["rustls"]
 openssl = ["actix-web/openssl", "awc/openssl"]
 rustls = ["actix-web/rustls", "awc/rustls"]
 shuttle = [
-  "dep:shuttle-actix-web", 
-  "dep:shuttle-runtime", 
-  "dep:shuttle-static-folder", 
+  "dep:shuttle-actix-web",
+  "dep:shuttle-runtime",
+  "dep:shuttle-static-folder",
   "dep:tokio"
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,12 @@ include = ["src/**/*", "Cargo.*", "LICENSE", "README.md", "CHANGELOG.md"]
 default = ["rustls"]
 openssl = ["actix-web/openssl", "awc/openssl"]
 rustls = ["actix-web/rustls", "awc/rustls"]
-shuttle = ["dep:shuttle-actix-web", "dep:shuttle-runtime", "dep:shuttle-static-folder", "dep:tokio"]
+shuttle = [
+  "dep:shuttle-actix-web", 
+  "dep:shuttle-runtime", 
+  "dep:shuttle-static-folder", 
+  "dep:tokio"
+]
 
 [dependencies]
 actix-web = { version = "4.3.1" }
@@ -25,7 +30,7 @@ actix-files = "0.6.2"
 awc = { version = "3.1.1" }
 env_logger = "0.10.0"
 log = "0.4.19"
-serde = "1.0.171"
+serde = "1.0.173"
 futures-util = "0.3.28"
 petname = { version = "1.1.3", default-features = false, features = [
   "std_rng",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ actix-files = "0.6.2"
 awc = { version = "3.1.1" }
 env_logger = "0.10.0"
 log = "0.4.19"
-serde = "1.0.173"
+serde = "1.0.174"
 futures-util = "0.3.28"
 petname = { version = "1.1.3", default-features = false, features = [
   "std_rng",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ actix-files = "0.6.2"
 awc = { version = "3.1.1" }
 env_logger = "0.10.0"
 log = "0.4.19"
-serde = "1.0.167"
+serde = "1.0.171"
 futures-util = "0.3.28"
 petname = { version = "1.1.3", default-features = false, features = [
   "std_rng",
@@ -35,17 +35,17 @@ rand = "0.8.5"
 dotenvy = "0.15.7"
 url = "2.4.0"
 mime = "0.3.17"
-regex = "1.9.0"
+regex = "1.9.1"
 serde_regex = "1.1.0"
-lazy-regex = "2.5.0"
+lazy-regex = "3.0.0"
 humantime = "2.1.0"
 humantime-serde = "1.1.1"
 glob = "0.3.1"
 ring = "0.16.20"
 hotwatch = "0.5.0"
-shuttle-actix-web = { version = "0.20.0", optional = true }
-shuttle-runtime = { version = "0.20.0", optional = true }
-shuttle-static-folder = { version = "0.20.0", optional = true }
+shuttle-actix-web = { version = "0.21.0", optional = true }
+shuttle-runtime = { version = "0.21.0", optional = true }
+shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
 serde_json = "1.0.100"
 
--- a/README.md
+++ b/README.md
@@ -255,6 +255,10 @@ $ rustypaste
 
 See [config.toml](./config.toml) for configuration options.
 
+### List endpoint
+
+Set `json_list_enabled` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
+
 #### HTML Form
 
 It is possible to use an HTML form for uploading files. To do so, you need to update two fields in your `config.toml`:
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ See [config.toml](./config.toml) for configuration options.
 
 ### List endpoint
 
-Set `json_list_enabled` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
+Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
 
 ```sh
 curl http://<server address>/list
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ See [config.toml](./config.toml) for configuration options.
 ### List endpoint
 
 Set `json_list_enabled` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
+This route will require an `AUTH_TOKEN` if one is set.
 
 #### HTML Form
 
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ $ curl -F "remote=https://example.com/file.png" "<server_address>"
 
 #### Cleaning up expired files
 
-Configure `delete_expired_files` to set an interval for deleting the expired files automatically.
+Configure `[paste].delete_expired_files` to set an interval for deleting the expired files automatically.
 
 On the other hand, following script can be used as [cron](https://en.wikipedia.org/wiki/Cron) for cleaning up the expired files manually:
 
@@ -253,6 +253,8 @@ $ echo "AUTH_TOKEN=$(openssl rand -base64 16)" > .env
 $ rustypaste
 ```
 
+You can also set multiple auth tokens via the array field `[server].auth_tokens` in your `config.toml`.
+
 See [config.toml](./config.toml) for configuration options.
 
 ### List endpoint
@@ -264,8 +266,8 @@ This route will require an `AUTH_TOKEN` if one is set.
 
 It is possible to use an HTML form for uploading files. To do so, you need to update two fields in your `config.toml`:
 
-- Set the `landing_page_content_type` to `text/html; charset=utf-8`.
-- Update the `landing_page` field with your HTML form.
+- Set the `[landing_page].content_type` to `text/html; charset=utf-8`.
+- Update the `[landing_page].text` field with your HTML form or point `[landing_page].file` to your html file.
 
 For an example, see [examples/html_form.toml](./examples/html_form.toml)
 
--- a/README.md
+++ b/README.md
@@ -262,7 +262,8 @@ See [config.toml](./config.toml) for configuration options.
 Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
 
 ```sh
-curl http://<server address>/list
+$ curl "http://<server_address>/list"
+
 [{"expires_at":null,"file_name":"accepted-cicada.txt","file_size":241}]
 ```
 This route will require an `AUTH_TOKEN` if one is set.
--- a/README.md
+++ b/README.md
@@ -267,6 +267,7 @@ $ curl "http://<server_address>/list"
 [{"expires_at":null,"file_name":"accepted-cicada.txt","file_size":241}]
 ```
 This route will require an `AUTH_TOKEN` if one is set.
+
 #### HTML Form
 
 It is possible to use an HTML form for uploading files. To do so, you need to update two fields in your `config.toml`:
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Here you can read the blog post about how it is deployed on Shuttle: [https://bl
     - [Paste file from remote URL](#paste-file-from-remote-url)
     - [Cleaning up expired files](#cleaning-up-expired-files)
   - [Server](#server)
+  - [List endpoint](#list-endpoint)
     - [HTML Form](#html-form)
     - [Docker](#docker)
     - [Nginx](#nginx)
@@ -259,13 +260,14 @@ See [config.toml](./config.toml) for configuration options.
 
 ### List endpoint
 
-Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls.
+Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot URLs, or URLs.
 
 ```sh
 $ curl "http://<server_address>/list"
 
 [{"file_name":"accepted-cicada.txt","file_size":241,"expires_at_utc":null}]
 ```
+
 This route will require an `AUTH_TOKEN` if one is set.
 
 #### HTML Form
--- a/README.md
+++ b/README.md
@@ -260,8 +260,12 @@ See [config.toml](./config.toml) for configuration options.
 ### List endpoint
 
 Set `json_list_enabled` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
-This route will require an `AUTH_TOKEN` if one is set.
 
+```sh
+curl http://<server address>/list
+[{"expires_at":null,"file_name":"accepted-cicada.txt","file_size":241}]
+```
+This route will require an `AUTH_TOKEN` if one is set.
 #### HTML Form
 
 It is possible to use an HTML form for uploading files. To do so, you need to update two fields in your `config.toml`:
--- a/README.md
+++ b/README.md
@@ -259,12 +259,12 @@ See [config.toml](./config.toml) for configuration options.
 
 ### List endpoint
 
-Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
+Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls.
 
 ```sh
 $ curl "http://<server_address>/list"
 
-[{"expires_at":null,"file_name":"accepted-cicada.txt","file_size":241}]
+[{"file_name":"accepted-cicada.txt","file_size":241,"expires_at_utc":null}]
 ```
 This route will require an `AUTH_TOKEN` if one is set.
 
--- a/config.toml
+++ b/config.toml
@@ -9,7 +9,7 @@ max_content_length = "10MB"
 upload_path = "./upload"
 timeout = "30s"
 expose_version = false
-expose_list = true
+expose_list = false
 #auth_tokens = [
 #  "super_secret_token1",
 #  "super_secret_token2",
--- a/config.toml
+++ b/config.toml
@@ -9,7 +9,7 @@ max_content_length = "10MB"
 upload_path = "./upload"
 timeout = "30s"
 expose_version = false
-json_list_enabled = true
+expose_list = true
 
 [landing_page]
 text = """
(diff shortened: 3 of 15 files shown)


Cargo.toml
```diff
@@ -40,18 +35,19 @@ rand = "0.8.5"
...
+lazy-regex = "3.0.0"
 humantime = "2.1.0"
 humantime-serde = "1.1.1"
 glob = "0.3.1"
 ring = "0.16.20"
 hotwatch = "0.5.0"
-shuttle-actix-web = { version = "0.20.0", optional = true }
-shuttle-runtime = { version = "0.20.0", optional = true }
-shuttle-static-folder = { version = "0.20.0", optional = true }
+shuttle-actix-web = { version = "0.21.0", optional = true }
```
ajbdev, line 46: I didn't intentionally mean to bump the version of these -- if they should remain the version they were on let me know and I will make sure to change them back.

tessus:
This looks great. I was thinking about implementing something similar. One suggestion though.

I think this should be guarded with an auth token. That is, if one is set. You can check how it was done for the `/version` and `/upload` endpoint.

ajbdev:
I'll implement the auth guard and make the change to the expiration variable (I'm not sure why I implemented it that way and agree that the expiration date definitely makes more sense!). Thanks for the feedback, let me know if there is anything else!

tessus:
I am just a user like yourself. I really only made suggestions, even though I think they made sense.
orhun is the owner of this project, so he is the one who makes all the decisions. ;-)

tessus:
Oh, I have just noticed that you are using the `/` endpoint for retrieving the list.

I am just playing the devil's advocate here. What would happen, if someone chose `version` as the `json_index_path`? Or any other endpoint that exists? I am not sure how actix handles that.

Maybe it would be better to use a hardcoded separate endpoint. Otherwise you migth have to check for collisions (make sure that no existing endpoints) are used.

Maybe orhun has some input on this as well.

Cargo.toml
orhun: Yup, please revert these changes back 🐻

README.md
```diff
@@ -264,6 +264,16 @@ It is possible to use an HTML form for uploading files. To do so, you need to up
 
 For an example, see [examples/html_form.toml](./examples/html_form.toml)
 
+#### Retrieving list of files
```
orhun, line 267: As @tessus pointed out, this should definitely be guarded behind an authentication token. Let's implement that and also make a note in README.md as well.

Also, see #28

orhun:
Hello, thanks for submitting this PR, it is definitely a nice feature to have!

> What would happen, if someone chose version as the json_index_path?

It is a valid concern. Having a hardcoded endpoint would be better in this case I think.

tessus:
@ajbdev if you go for a hardcoded endpoint, something simple and short like `/list` should do. ;-)

Btw, when you are done with your frontend, I'd be interested in checking it out in case you are going to make it public on github.

ajbdev:
This all sounds good - I will change to the hardcoded endpoint! Yes, I'll be sharing my frontend publicly, although it won't be a very traditional pastebin UI most likely.

tessus:
> although it won't be a very traditional pastebin UI most likely.

For me this is perfectly fine. I am just interested in what kind of frontend you are cooking up. I mostly use the command line, but now and then a UI (text or graphics based) can be useful.
I am also interested in this, because I really suck in anything frontend like. Haha.

Cargo.toml
ajbdev: Ok, they should be reverted to only the necessary changes now.

ajbdev:
I had some time to make these changes and add test coverage. Let me know if I missed anything or you see anything else to change.

Cargo.toml
```diff
@@ -52,6 +52,8 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
+serde_json = "1.0.100"
```
tessus, line 55:
```suggestion
serde_json = "1.0.103"
```

Cargo.toml
```diff
@@ -30,7 +30,7 @@ actix-files = "0.6.2"
 awc = { version = "3.1.1" }
 env_logger = "0.10.0"
 log = "0.4.19"
-serde = "1.0.171"
+serde = "1.0.173"
```
tessus, line 33:
```suggestion
serde = "1.0.174"
```

Cargo.toml
```diff
@@ -17,10 +17,10 @@ default = ["rustls"]
 openssl = ["actix-web/openssl", "awc/openssl"]
 rustls = ["actix-web/rustls", "awc/rustls"]
 shuttle = [
-  "dep:shuttle-actix-web",
-  "dep:shuttle-runtime",
-  "dep:shuttle-static-folder",
-  "dep:tokio",
+  "dep:shuttle-actix-web", 
+  "dep:shuttle-runtime", 
+  "dep:shuttle-static-folder",
```
tessus, line 22: why the space after the comma?

tessus (commented):
LGTM, but I am probably not qualified to review Rust code. There are many constructs in Rust I just don't know about, but make code more concise.
Putting aside the fact that I don't know whether there are better ways to write the code, the code is sound.

I also had a few inline comments, none of which are really of any importance.

In case my PR gets merged first, it will be necessary to do a minor change to the auth check call. But I will comment inline when this happens.

tessus:
I also like it that you added a config option to enable/disable the feature. 👍

src/server.rs
```diff
@@ -297,10 +302,81 @@ async fn upload(
...
+    request: HttpRequest,
+    config: web::Data<RwLock<Config>>,
+) -> Result<HttpResponse, Error> {
+    let config = config
+        .read()
+        .map_err(|_| error::ErrorInternalServerError("cannot acquire config"))?
+        .clone();
+
+    if !config.server.json_list_enabled.unwrap_or(false) {
+        Err(error::ErrorForbidden("json list is not enabled"))?;
```
tessus, line 324: maybe this should more consistent with the `/version` endpoint.

e.g. adding a warning 

```suggestion
        log::warn!("server is not configured to expose list endpoint");
Err(error::ErrorForbidden("endpoint is not exposed"))?;
```

But it's just an idea.

src/server.rs
```diff
@@ -297,10 +302,81 @@ async fn upload(
...
+    let connection = request.connection_info().clone();
+    let host = connection.realip_remote_addr().unwrap_or("unknown host");
+
+    auth::check(
+        host,
+        request.headers(),
+        env::var(AUTH_TOKEN_ENV)
+            .ok()
+            .or_else(|| config.server.auth_token.as_ref().cloned()),
+    )?;
```
tessus, line 336: Here is the change for the auth call.

```suggestion
     let tokens = config.get_tokens();
     auth::check(host, request.headers(), tokens)?;
```

Cargo.toml
orhun: yeah, please revert the changes in `Cargo.*`

Cargo.toml
```diff
@@ -52,6 +52,8 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
+serde_json = "1.0.100"
+chrono = "0.4.26"
```
orhun, line 56: Let's remove the `chrono` dependency and deliver the UNIX timestamp instead of formatted time.

README.md
```diff
@@ -255,6 +255,11 @@ $ rustypaste
 
 See [config.toml](./config.toml) for configuration options.
 
+### List endpoint
+
+Set `json_list_enabled` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls.
```
orhun, line 260: Just brainstorming, should we maybe rename this to `list_files` or something similar?

```diff
- json_list_enabled = true
+ list_files = true
```

Also, I think there is no point in prefixing it with `json` since it is a no-brainer to use JSON 😅 

What do you think? @tessus @ajbdev

README.md
```diff
@@ -255,6 +255,11 @@ $ rustypaste
 
 See [config.toml](./config.toml) for configuration options.
 
+### List endpoint
+
+Set `json_list_enabled` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
+This route will require an `AUTH_TOKEN` if one is set.
+
```
orhun, line 262: Can we also give an example of the returned JSON here?

shuttle/config.toml
```diff
@@ -10,6 +10,7 @@ upload_path = "./upload"
 timeout = "30s"
 expose_version = true
 landing_page = """
+
```
orhun, line 13: Please remove this unnecessary newline.

src/server.rs
```diff
@@ -4,19 +4,22 @@ use crate::file::Directory;
 use crate::header::{self, ContentDisposition};
 use crate::mime as mime_util;
 use crate::paste::{Paste, PasteType};
-use crate::util;
+use crate::util::{self};
```
orhun, line 7: nit: Unnecessary import.

src/server.rs
```diff
@@ -297,10 +302,81 @@ async fn upload(
     Ok(HttpResponse::Ok().body(urls.join("")))
 }
 
+#[derive(Serialize, Deserialize)]
+struct ListItem {
+    file_name: String,
+    file_size: u64,
+    expires_at: Option<String>,
+}
```
orhun, line 310: Struct fields are not documented. Also, consider making this `pub`.

src/server.rs
orhun: Agree with @tessus here.

src/server.rs
```diff
@@ -297,10 +302,81 @@ async fn upload(
...
+                    .and_then(|v| v.parse().ok());
+
+                let mut expires_at = None;
+
+                if let Some(expiration) = extension {
+                    let seconds = expiration / 1000;
+                    let dt = NaiveDateTime::parse_from_str(&seconds.to_string(), "%s")
+                        .expect("invalid unix timestamp");
+
+                    expires_at = Some(dt.format("%Y-%m-%d %H:%M:%S").to_string());
```
orhun, line 360: Let's use the UNIX timestamp instead of formatting the time here.

src/server.rs
```diff
@@ -523,6 +599,28 @@ mod tests {
...
+                .configure(configure_routes),
+        )
+        .await;
+
+        let request = TestRequest::default()
+            .insert_header(("content-type", "text/plain"))
+            .uri("/list")
+            .to_request();
+
+        let _result: Vec<ListItem> = test::call_and_read_body_json(&app, request).await;
```
orhun, line 619: Result is not checked.

orhun (changes requested):
Thanks a lot for your contribution! I had some comments.

(also thanks @tessus for the review!)

Cargo.toml
tessus: I do understand where you are coming from, but if I am the admin of the instance (or someone who uploaded a bunch of files), I'd like to see when the files expire. I am not bad in mathematics, but I can't do that calculation quickly in my head.
Nor do I know off the top of my head whether I can use a bash function in jq to convert a field to another output, otherwise `date +'%Y-%m-%d %H:%M:%S' -d "@unixtimestamp"` would do the trick.

Maybe we could have 2 fields: e.g. `expiry_date` and `expiry_date_human` or something similar. 

Otherwise I would have to write a tool/wrapper to convert the unix timestamps. 

This is easy for people who use rpaste where a flag can be implemented to show human readable timestamps, but when someone uses curl, the unix timestamps are rather cumbersome.

README.md
tessus: Short answer: yes.

Long answer: If it were to be consistent, it should be `expose_list`. But then I went down the rabbit hole.... What about a var e.g. `expose = ['version', 'list']`? In that case, another deprecation would be necessary. Although using an array to expose endpoints is more future proof. But maybe too much for the scope of this PR. (Just blabbing out thoughts unfiltered.)

Cargo.toml
ajbdev: If the goal is to remove the dependency on chrono, I think it might be possible to do this using PrimitiveDateTime from the std time lib. 

I'll proceed with any approach you guys agree on.

README.md
ajbdev: I like `expose_list`

Cargo.toml
ajbdev: I was developing on windows on top of git bash and it seemed to do weird things with the formatting with my vscode settings. Sorry about that. I think the package version changes below are still needed though.

src/server.rs
```diff
@@ -283,10 +288,79 @@ async fn upload(
...
+        .clone();
+    let connection = request.connection_info().clone();
+    let host = connection.realip_remote_addr().unwrap_or("unknown host");
+
+    let tokens = config.get_tokens();
+    auth::check(host, request.headers(), tokens)?;
+
+    if !config.server.expose_list.unwrap_or(false) {
+        log::warn!("server is not configured to expose list endpoint");
+        Err(error::ErrorForbidden("endpoint is not exposed"))?;
```
tessus, line 320:
```suggestion
        Err(error::ErrorForbidden("endpoint is not exposed\n"))?;
```

README.md
orhun: Yup, let's go with `expose_list`. Sounds reasonable.

Cargo.toml
orhun: If it is possible to format the time without `chrono`, let's go with that then.

tessus: May I ask why chrono is bad? Is it too big or does it have security issues? I am not trying to be difficult, I really don't know.

orhun: There was a security issue for a long time and people have been switching to the `time` crate.

- https://rustsec.org/advisories/RUSTSEC-2020-0159.html
- https://stackoverflow.com/questions/73883691/the-latest-chrono-0-4-crate-uses-time-0-1-which-has-a-potential-segfault-how-t

Also, if it is just for formatting, there is no need to pull in a lot of dependencies I think.

tessus: I did some research and it is nearly impossible to get a human readable date/time string with the Rust standard library.

Well, that's not entirely true. One has to write the code themselves. e.g. this is the [algorithm](http://git.musl-libc.org/cgit/musl/tree/src/time/__secs_to_tm.c?h=v0.9.15) that MUSL uses.

I can't believe this is not a part of Rust. One really does not need all the fancy stuff these libs provide. A simple function that returns a struct of 6 vars or just the vars themselves (year, month, day, hour, minute, second) would be enough. Maybe I should write a crate that does exactly that.

tessus: I am seriously contemplating whether I should write a crate. It would have exactly one function which would take an i64 as an argument and return a struct with 6 members (Option<struct>).

tessus: I am almost done. The only thing missing is the README and the doc comments.
I'll do that this evening and publish the crate.

tessus: There you go: https://crates.io/crates/uts2ts

The docs need a while before they are generated, but I am pretty sure the example and README already explain everything.

ajbdev: Nice work on this, it was very easy to implement.

orhun: Wow, @tessus just dropped a new crate outta nowhere. Nice! 🫡

README.md
```diff
@@ -257,6 +257,15 @@ You can also set multiple auth tokens via the array field `[server].auth_tokens`
 
 See [config.toml](./config.toml) for configuration options.
 
+### List endpoint
+
+Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot urls, or urls. 
+
+```sh
+curl http://<server address>/list
+[{"expires_at":null,"file_name":"accepted-cicada.txt","file_size":241}]
```
orhun, line 266:
```suggestion
$ curl "http://<server_address>/list"

[{"expires_at":null,"file_name":"accepted-cicada.txt","file_size":241}]
```

README.md
```diff
@@ -257,6 +257,15 @@ You can also set multiple auth tokens via the array field `[server].auth_tokens`
...
+```
+This route will require an `AUTH_TOKEN` if one is set.
```
orhun, line 268:
```suggestion
This route will require an `AUTH_TOKEN` if one is set.

```

config.toml
```diff
@@ -9,6 +9,7 @@ max_content_length = "10MB"
 upload_path = "./upload"
 timeout = "30s"
 expose_version = false
+expose_list = true
```
orhun, line 12: Let's disable this feature since the auth tokens are not set as default.

src/config.rs
```diff
@@ -58,6 +58,8 @@ pub struct ServerConfig {
     /// Landing page content-type.
     #[deprecated(note = "use the [landing_page] table")]
     pub landing_page_content_type: Option<String>,
+    /// Set the path for the JSON index
```
orhun, line 61:
```suggestion
    /// Path of the JSON index.
```

src/server.rs
```diff
@@ -283,10 +286,78 @@ async fn upload(
     Ok(HttpResponse::Ok().body(urls.join("")))
 }
 
+/// File entry item for list endpoint
```
orhun, line 289:
```suggestion
/// File entry item for list endpoint.
```

src/server.rs
```diff
@@ -283,10 +286,78 @@ async fn upload(
...
+#[derive(Serialize, Deserialize)]
+struct ListItem {
+    /// Uploaded file name
+    pub file_name: String,
+    /// Size of the file in bytes
+    pub file_size: u64,
+    /// ISO8601 formatted date-time string of the expiration timestamp if one exists for this file.
+    pub expires_at: Option<String>,
+}
```
orhun, line 298:
```suggestion
#[derive(Serialize, Deserialize)]
pub struct ListItem {
    /// Uploaded file name.
    pub file_name: String,
    /// Size of the file in bytes.
    pub file_size: u64,
    /// ISO8601 formatted date-time string of the expiration timestamp if one exists for this file.
    pub expires_at: Option<String>,
}
```

src/server.rs
```diff
@@ -283,10 +286,78 @@ async fn upload(
...
+
+/// Returns a list of files
```
orhun, line 300:
```suggestion
/// Returns the list of files.
```

src/server.rs
```diff
@@ -513,6 +584,45 @@ mod tests {
...
+        .await;
+
+        let request = TestRequest::default()
+            .insert_header(("content-type", "text/plain"))
+            .uri("/list")
+            .to_request();
+        let result: Vec<ListItem> = test::call_and_read_body_json(&app, request).await;
+
+        assert_eq!(result.len(), 1);
+        assert_eq!(result.first().unwrap().file_name, file_name);
```
orhun, line 619: Remove `unwrap()` for fixing clippy.

Cargo.toml
```diff
@@ -52,6 +52,8 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
+serde_json = "1.0.103"
```
orhun, line 55: This dependency is not needed, please remove it.

src/server.rs
```diff
@@ -12,11 +12,14 @@ use awc::Client;
 use byte_unit::Byte;
 use futures_util::stream::StreamExt;
 use mime::TEXT_PLAIN_UTF_8;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
```
orhun, line 16:
```suggestion

```

src/server.rs
```diff
@@ -283,10 +286,78 @@ async fn upload(
...
+                Some(ListItem {
+                    file_name,
+                    file_size: metadata.len(),
+                    expires_at,
+                })
+            })
+        })
+        .collect();
+
+    Ok(HttpResponse::Ok().json(json!(entries)))
```
orhun, line 353:
```suggestion    
    Ok(HttpResponse::Ok().json(entries))
```

orhun (changes requested):

Cargo.toml
tessus: Thanks guys! I was inspired by this situation. I really got annoyed that such a simple thing was missing from Rust. Pulling in `chrono` or `time` seemed like an overkill. 

I made sure that it will work for dates after 2038. With a 64bit implementation the date range is 584 billion years. I think that should do it. Which is why I did not bother to allow input of time units (ms, us, ns) other than seconds.

This crate does not even do the reverse. If something like that was needed, I'd write another crate `ts2uts` ;-) haha.

It gave me the opportunity to learn more about the Rust ecosystem, even though I still need to figure out a few things to improve my workflow. On one side there is a gh action workflow that does the deployment for me. But I still have to bump the version and change files manually. I saw that there are things like cargo-release and other tools, but I need them to work together. Testing all this stuff on a live project seems wrong.

If you know a nice hook manager, please let me know. There are a bunch of nice ones out there, but I haven't settled on one yet.

Btw, I stole your PR request template, but mentioned this most despicable deed in the commit message https://github.com/tessus/uts2ts/commit/0f604f9ee9b015fd056cd7d462f30e1a373ba556 ;-)

tessus: I am thinking of changing the struct members (except year) to `u8`. This will not have any impact on anything, but the struct will allocate less memory. What do you think?

src/server.rs
```diff
@@ -513,6 +583,45 @@ mod tests {
...
+        .await;
+
+        let request = TestRequest::default()
+            .insert_header(("content-type", "text/plain"))
+            .uri("/list")
+            .to_request();
+        let result: Vec<ListItem> = test::call_and_read_body_json(&app, request).await;
+
+        assert_eq!(result.len(), 1);
+        assert!(result.first().is_some());
```
tessus, line 618: assert_eq!(result.first().expect("json object").file_name, file_name);
```suggestion
        assert!(result.first().is_some());
```

A panic is perfectly fine here. `assert` also panics, if the condition is not met.

The issue was that clippy complained about unwrap, which in my opinion should be ignored by clippy for tests.

To get rid of the message there are 2 options:

- allow unwrap for clippy for that test
- use `expect` instead

Cargo.toml
```diff
@@ -52,6 +52,7 @@ shuttle-actix-web = { version = "0.21.0", optional = true }
 shuttle-runtime = { version = "0.21.0", optional = true }
 shuttle-static-folder = { version = "0.21.0", optional = true }
 tokio = { version = "1.29.1", optional = true }
+uts2ts = "0.1.0"
```
tessus, line 55:
```suggestion
uts2ts = "0.3.0"
```

I have changed the crate so that less memory is allocated for the struct. It's probably not a big win in regards to perf, but it doesn't hurt either.

src/server.rs
```diff
@@ -283,10 +285,78 @@ async fn upload(
     Ok(HttpResponse::Ok().body(urls.join("")))
 }
 
+/// File entry item for list endpoint.
+#[derive(Serialize, Deserialize)]
+pub struct ListItem {
+    /// Uploaded file name.
+    pub file_name: String,
+    /// Size of the file in bytes.
+    pub file_size: u64,
```
tessus, line 294: I am wondering, whether the underscore is necessay for the first 2 items.

Isn't `filename` and `filesize` rather used in most situations?

src/server.rs
```diff
@@ -283,10 +285,78 @@ async fn upload(
...
+    /// ISO8601 formatted date-time string of the expiration timestamp if one exists for this file.
+    pub expires_at: Option<String>,
```
tessus, line 296: Sorry for my question, but isn't a noun better? e.g. `expiry` or `expiration`

I am not sure about that one, but I believe I have seen more nouns as keys than verbs.

src/server.rs
ajbdev: `file_name` gets used pretty frequently elsewhere in this source but it doesn't matter to me either way.

src/server.rs
ajbdev: I usually use `_at` in my timestamps so it can be known that the field is a timestamp just by reading the name, but that's mostly a holdover from my days as a Rails programmer (where it is a convention). That reasoning makes less sense in a typed language, though. I personally am not dogmatic about using nouns as property fields, I tend to use whatever I think makes the meaning most clear. I'm fine changing it though, especially if there is already a convention being used to follow.

src/server.rs
ajbdev: Thank you! This is where my lack of Rust knowledge bites me... I think the `expect` syntax is simple enough.

src/server.rs
tessus: right, but in this case it is public and the key name in the JSON output. e.g. if you were to use jq or any other tool to filter the output, IMO keys without a underscore would be easier to type. And fom filename looks better than file_name.
But that's subjective. I am just lazy when it comes to typing.

src/server.rs
tessus: No need to change it. It was just something that came to mind.

src/server.rs
tessus: Same here. I am always happy when orhun shows me the Rust way of doing things...

src/server.rs
ajbdev: Sounds logical to me. I went ahead and made this change.

ajbdev:
These commits are pretty noisey, I can rebase them if you'd like just let me know (unless you'll just squash merge anyways)

tessus:
orhun squashes them when he merges the PR. I haven't seen any indication that during the lifetime of a PR he wanted to squash and force-push, but I might be wrong.

tessus:
I just noticed that the filename also includes the timestamp:

```json
[
  {
    "filename": "foo.rjOq.bar.txt.1690750967938",
    "filesize": 12,
    "expires_at": "2023-07-30 21:02:47"
  }
]
```

Is this on purpose?

When you create a link BASE_URL + filename, the result (in a browser) is always that the file is downloaded, because the mimetype is not recognized with this (timestamp) extension.

For me this is not an issue, but might not be as intended. Maybe the timestamp should be moved to its own separate field. But I really do not know, and even if I were the person to decide, I still couldn't make the decision alone, because I am not sure what the best solution is. There are a few options and all of them have their pros/cons:

- keep it as is
- remove the timestamp from the filename and do nothing else
- move the timestamp from the filename to a separate field 

What do you guys think @orhun and @ajbdev?

ajbdev:
I think we should just chop it off. The expiration is already a separate field.

orhun:
> I think we should just chop it off. The expiration is already a separate field.

Yup, let's do that.

> The easiest and most performant is to just add it to the key.

Agreed!

src/server.rs
```diff
@@ -283,10 +285,78 @@ async fn upload(
...
+) -> Result<HttpResponse, Error> {
+    let config = config
+        .read()
+        .map_err(|_| error::ErrorInternalServerError("cannot acquire config"))?
+        .clone();
+    let connection = request.connection_info().clone();
+    let host = connection.realip_remote_addr().unwrap_or("unknown host");
+
+    let tokens = config.get_tokens();
+    auth::check(host, request.headers(), tokens)?;
```
orhun, line 313: In order to _disable_ this feature when no authorization is set, you need to check the existence of `tokens` (and return an error such as "authorization is not set") instead of the normal authentication flow.

Right now the files are served if `expose_list` is set to `true`, regardless of authentication tokens.

src/server.rs
```diff
@@ -283,10 +285,78 @@ async fn upload(
     Ok(HttpResponse::Ok().body(urls.join("")))
 }
 
+/// File entry item for list endpoint.
+#[derive(Serialize, Deserialize)]
+pub struct ListItem {
+    /// Uploaded file name.
+    pub filename: String,
+    /// Size of the file in bytes.
+    pub filesize: u64,
```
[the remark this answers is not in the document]
orhun: Just went ahead and did this change. Also added fixture tests in f7c0fc91946d1017f5a70a92573cf04d32c74f25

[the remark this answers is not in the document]
tessus: Any chance to use rename then? I really haven't seen the term `file_name` in userland in Unix/Linux. man pages, documentation, UI elements, ....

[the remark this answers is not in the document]
orhun: > Any chance to use rename then?

Do you mean `#[serde(rename)]`?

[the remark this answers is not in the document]
tessus: Yes, I meant serde rename. But it is really not important. I just wanted to point out that `file_name` and `file_size` is not very common when it is customer facing. At least I can't recall any docs that use underscore for these. it's either: file name, filename, file's name, but hardly file_name. But who am I to say that the current status quo won't change, nor that I am not mistaken. Maybe I just haven't read enough docs.

I am perfectly fine with file_name and file_size.

src/server.rs
tessus: Isn't this is the same flow as for `/version`?

I also tested this manually and it works as expected.

ajbdev:
Hey guys, I understand this PR may need further changes but for the purposes that I need its more than adequate. Unfortunately (with a newborn to take care of) I can only spend so much time tending to hobby projects, so if you'd like to merge the PR someone else will have to implement further changes.  As much as I'd love to know how to be a competent rustacean life sometimes gets in the way :) I hope you all understand.

orhun:
Congrats @ajbdev! Thank you for your efforts so far and definitely no worries, we will be finishing this up with @tessus for sure!

src/server.rs
tessus: I think this is pretty much the last item to clarify. Or is there anything else?

Maybe I misunderstood your concern, but I can't seem to find an issue with the current implementation. I also listed the test cases above. I haven't uploaded any files, thus the output for a successful request/respones is just `[]`. (At the beginning of the line, since there's no newline at the end of the json output.)

tessus:
> > The easiest and most performant is to just add it to the key.
> 
> Agreed!

I also thought about maybe adding one more method to the crate `as_string_utc()`. Because in reality with this kind of primitive transformation without any TZ awareness there are only 2 options to present a date/time to the user:

- local (whatever that might be) without TZ info
- UTC (also showing that it is in fact UTC)

No harm in adding the second one of these 2 options to the crate.

tessus:
@orhun I think this should be ready. please also have a look at https://github.com/orhun/rustypaste/pull/94#discussion_r1279530585

src/server.rs
orhun: I meant disabling the `list` endpoint completely when no tokens is set but with a second thought, it would be better if we follow the same logic as `version` endpoint.

orhun:
I cleaned up the implementation in b9e15dcf5c965177bacb1798af053727f01b6469 (it might be worth checking out)

orhun (approved):
Thank you both @ajbdev @tessus for your efforts 💖

orhun:
P.S. Would you be interested in adding this feature to [rustypaste-cli](https://github.com/orhun/rustypaste-cli) as well?

tessus:
> I cleaned up the implementation in [b9e15dc](https://github.com/orhun/rustypaste/commit/b9e15dcf5c965177bacb1798af053727f01b6469) (it might be worth checking out)

Nice refactor indeed.

> Would you be interested in adding this feature to [rustypaste-cli](https://github.com/orhun/rustypaste-cli) as well?

Yes, I was thinking about it, but was hung up on how to present the data. When I thought about the formatting, a few options came to mind. The best would probably be a table, but I'd have to iterate through the json to find the longest filename and the largest filesize to determine the width of the columns. This also means iterating through the JSON twice.
Or the first column could be the expiry timestamp, the second column the file size, and the third column the file name. Then the column widths could be static.

Then how about the file size? Bytes or units (kb, mb, gb, tb, ...). e.g. I use the term MB, but calculate with 1024. Many, many years back they renamed the MB that is binary based to MiB (which I really hate). I grew up with learning that in IT KB/MB/GB are not factors of 1000 but 1024.
Then these idiot hard disk manufacturers started to calculate with 1000 and people were angry and confused that they got less storage than they expected. As far as I know RAM manufacturers still use the term MB but caluclate with 1024.

Sorry for my rant, but this stuff still makes me angry every time I think about it.

tessus:
While working on the cli code (almost done, btw), I have noticed that expired files are also listed. There should be a check (on the server) inverse to what is done when removing expired files when building the list of files.

orhun:
> Then these idiot hard disk manufacturers started to calculate with 1000 and people were angry and confused that they got less storage than they expected. As far as I know RAM manufacturers still use the term MB but caluclate with 1024.

Lol, yeah, I can totally relate with this rant 😄

PR #107, opened 2023-08-02
feat(server): support handling spaces in filenames

omerbustun:
## Description
Added an option to enable encoding for returned URL. It can be enabled in the config by setting `url_encode_filenames = true`.

## Motivation and Context
My terminal wasn't hyperlinking the returned URL when the file name had spaces in it. I wanted to be able to click on the URL in the terminal rather than copying and pasting it in a browser. 

## How Has This Been Tested?
- Test cases
- Manual testing

### Added

- Add a function that encodes the file name and consequently the returned URL.
- Add a config option to turn the feature on and off.

## Types of Changes

- [ ] Other 

## Checklist:


orhun:
Hello, can you give an use case example for this?

orhun:
Can you also push the updated `Cargo.lock`?

orhun:
ping @omerbustun

omerbustun:
Hello @tessus,

Yes, the [`urlencoding`](https://docs.rs/urlencoding/latest/urlencoding/) crate I utilized in the feature handles the encoding of unicode characters seamlessly. It assumes UTF-8 encoding and percent-encodes every byte except alphanumerics and some special characters (`-`, `_`, `.` and `~`).

As per the examples you provided, they become the following:

```
https://paste.site.com/%EA%B0%90%EC%82%AC%ED%95%A9%EB%8B%88%EB%8B%A4
https://paste.site.com/%F0%9F%A4%AF.txt
```

I hope this clarifies the behavior of the encoding with respect to unicode characters.

Best regards,
Ömer

tessus:
Thanks for the info. It's just that IMO unicode characters do not need to be encoded. Most terminals are unicode compatible and reading an encoded unicode string is impossible. e.g I don't know that `%EA%B0%90%EC%82%AC%ED%95%A9%EB%8B%88%EB%8B%A4` is `감사합니다`.
This might be ok for people who write mostly in the ascii alphabet, but everything else is unreadable.

I think the only reason you want to do this encoding is to be able to click on a link with spaces. Woudn't it be better to only encode the spaces (when random URLs off)? 
To be fair, clients all handle spaces and unicode. You can curl unicode URLs (and URLs with spaces) or paste them in the browser.

Or maybe rustypaste should just remove spaces from the file name.

omerbustun:
You've got a point about modern clients handling spaces and unicode in URLs without any issues. However, the primary motivation behind my proposal was rooted more in user convenience than technical compatibility. In my experience with Windows Terminal, URLs with spaces aren't recognized as clickable links, as shown in the screenshot.

resim

Your suggestion to only encode spaces or replace them with underscores is a viable alternative that would address the main concern while preserving the readability of filenames with unicode characters. This way, non-ASCII users won't be faced with an unreadable string, yet the core functionality of having clickable links in terminals would still be achieved.

It's also worth noting that since this feature can be toggled on or off via the config, users who prefer the original behavior or find the encoding unnecessary can simply keep it disabled.

With that said, I'm open to refining this feature based on your feedback. Would you prefer the encoding to be limited to spaces only, or to replace spaces with underscores? I'm flexible and willing to adjust the implementation accordingly.

Best,
Ömer

tessus:
I was just thinking out loud. Orhun will have to decide how to handle this.

Since I almost never use file names with spaces, I don't really have this problem. I would just hate to see such an encoding as the default and only behavior. If it were behind a config option (as you have coded it), I wouldn't really care.

But in reality, how many people deactivate the random feature just to keep the file name as is? Don't forget, this also has the disadvantage that files with the same name are overwritten.

Personally I think that replacing spaces with underscores is the best option, but that's just my opinion.

oflifurkan:
> I was just thinking out loud. Orhun will have to decide how to handle this.
> 
> Since I almost never use file names with spaces, I don't really have this problem. I would just hate to see such an encoding as the default and only behavior. If it were behind a config option (as you have coded it), I wouldn't really care.
> 
> But in reality, how many people deactivate the random feature just to keep the file name as is? Don't forget, this also has the disadvantage that files with the same name are overwritten.
> 
> Personally I think that replacing spaces with underscores is the best option, but that's just my opinion.

As someone who uses this project against my friends and family in my daily life, I can say that I think this pull request is very necessary for the direct copying of the links thrown on social media applications on mobile. In mobile programs, it can cause problems because there is no text selection feature in the message bubbles.
I think that the ability to encode url as an option in this pull request is a well-thought-out feature of this PR.

omerbustun:
> But in reality, how many people deactivate the random feature just to keep the file name as is? Don't forget, this also has the disadvantage that files with the same name are overwritten.


In my setup, I use the suffix_mode for the random URL. This mode allows me to retain a part of the original filename while adding a unique suffix. So I can differentiate the files by their URL and also ensure that files with the same name don't get overwritten.

tessus:
> As a regular user of this project for my friends and family, I've noticed a recurring issue when sharing links with spaces. When I copy and send these links through apps like Discord and WhatsApp, they aren't recognized as hyperlinks due to the spaces.

Which can also be solved by replacing spaces with underscores.

Don't get me wrong, I am not against this PR or making links with spaces clickable. This is not what I am saying. I am just asking myself if adding another dependency for encoding URLs is necessary when the same can be achieved by just replacing spaces. That's all.

orhun:
It is an edge case but I think it doesn't hurt to have this feature.

But before merge, @omerbustun can you complete the following checklist?

- [ ] Resolve conflicts with `master`
- [ ] Add unit tests for this configuration option
- [ ] Add fixture tests (see [fixtures](https://github.com/orhun/rustypaste/tree/master/fixtures))

tessus:
Sorry for being a pain in the ass, but wouldn't it be much easier and less complex in terms of future maintenance overhead just to add one line to `paste.rs`?

```rust
        let file_name = path
            .file_name()
            .map(|v| v.to_string_lossy())
            .unwrap_or_default()
            .to_string()
            .replace(" ", "_");                // <---- this line here on line 169 
```

orhun:
@tessus actually yeah, it seems like the only point of using URL encoding is to replace the whitespaces with underscores. I personally never hit a case like this so I'm okay with both ways. Of course, avoiding a dependency would be nice considering that `urlencoding` is now [unmaintained](https://github.com/kornelski/rust_urlencoding/commit/3cc277fc4e577a0bef99294d7ab9bc85232f0730).

tessus:
@omerbustun may I ask why the code is more than the one line I suggested? The filename is not completely retained anyway, because a random suffix is added. Or do you really want to keep the spaces in the filename? rustypaste is not a storage backend, accessed via an NFS or Samba share. I am not against it, I am just curious.

For me translating spaces into underscores would not even have to be guarded by a config option. I think rustypaste is for quickly making data available to others. There is no need to retain a literal space in the URL. Or is there?

omerbustun:
Hi @orhun and @tessus,

I've made some modifications to the approach based on the feedback:

 - **Encoding whitespaces:** Instead of using the urlencoding library to encode all characters, I've now simplified the approach to only replace whitespaces with %20. This directly addresses the primary concern of filenames with spaces, and keeps the implementation straightforward.
 - **Tests:** I've included both unit and fixture tests as you requested, @orhun.
- **Dependency on urlencoding:** While I've removed the direct use of the urlencoding library in my approach, it's worth noting that this library is already a transitive dependency in the project, brought in by other dependencies.
- **Regarding the suggested change:** @tessus, while replacing spaces with underscores is a simpler approach, the advantage of the configuration option I proposed is flexibility. Users who find the feature unnecessary can keep it turned off, while others who need it can enable it. Your approach, although simpler, would always replace spaces, making it a global change. In the end it all boils down to preference, I prefer having and providing options.

tessus:
No worries, I just meant it's not a bad idea to replace spaces with underscores globally. File names are not retained anyway, so why would I care, whether there is a space or an underscore?

When I added the random suffix mode, I only wanted to make sure that people (including myself) already have an idea of what could be in the file. Which is absolutely impossible to ascertain by reading a file name called `hopping-elephant.log`. ;-)

For me a file name is easier to read with underscores than with `%20`. That's just my opinion, which is of course subjective.

As mwntioned before, I am not against this PR or the config parameter.

orhun:
Thanks for the changes @omerbustun!

> For me a file name is easier to read with underscores than with %20.

I would say yes, I would vote against using `%20` instead of `_` which made me think about the following idea.

---

Right now we have this:

```toml
[server]
url_encode_filenames = true
```

Instead, I propose:

```toml
[paste]
replace_filename = [ { from = " ", to = "%20" } ]
```

So that you can configure it to replace whitespaces with underscores as follows:

```toml
[paste]
replace_filename = [ { from = " ", to = "_" } ]
```

Or even multiple characters/strings:

```toml
[paste]
replace_filename = [ { from = " ", to = "---" }, { from = "foo", to = "bar" } ]
```

What do you think? @omerbustun @tessus

tessus:
I think the issue is a bit more complicated.

The current implementation does not change the filename in the `upload` directory. It only returns the filename encoded with `%20` instead of spaces.

My idea of just replacing spaces with underscores would take place in `paste.rs` and thus the file would not be stored with spaces in the `upload` dir, but with underscores. Which is also fine, but the concept is different.

The idea of a mapping function is intriguing, but introduces a certain complexity to the code:

- when will the mapping be applied
  - before/after adding the suffix
  - before/after dding the extension
- are we storing the file in the `upload` dir with `%20`?
- certain character must be filtered out
  - from: `.` should not be replaced
  - to: `.`, `/`, `\`, `:` have certain meanings in filesystems and must be avoided in a filename
- there might be an attack vector in such a replacement, unless a thorough filter and checking of user input is done

While I really like the idea, it's a huge maintenance burden and adds unnecessary complexity to the code.

IMO, there are 3 options:

- leave it in the current state of this PR
- do the space to underscore replacement (w/o a config option)
- rename the config option to `paste.replace_spaces` and allow only 2 options:
  - `percent` (which replaces the spaces with `%20`)
  - `underscore` (which replaces the spaces with `_`)

omerbustun:
Hi @orhun and @tessus,

Thank you both for the active discussion. Here's my two cents.

- The PR encodes the returned URL, not the filename. The filename in the upload directory and when downloaded retains its spaces. URLs need to be hyperlink-friendly; filenames don't have to be. Hardcoding `%20` into filenames isn't the objective.
- Replacing spaces with underscores in filenames might be practical for URLs but moves away from this PR's intent. If users disable `[random_url]`, they might want original filenames intact.
- `replace_filename` is an interesting proposal, but it does introduce complexity and potential edge cases as pointed out by @tessus.


Given the above, this PR's approach provides a balance between usability and preserving original file names so I'm leaning towards leaving it as is in the options stated by @tessus. Appreciate the feedback, and let's determine the best direction.

omerbustun:
Hi @orhun and @tessus,

Thank you both for the active discussion. Here's my two cents.

- The PR encodes the URL, not the filename. The file in the upload directory and when downloaded retains its spaces. URLs need to be hyperlink-friendly; filenames don't have to be. Hardcoding `%20` into filenames isn't the objective.
- Replacing spaces with underscores in filenames might be practical for URLs but moves away from this PR's intent. If users disable `random_url`, they might want original filenames intact.
- `replace_filename` is an interesting proposal, but it does introduce complexity and potential edge cases as pointed out by @tessus.


Given the above, this PR's approach provides a balance between usability and preserving original file names so I'm leaning towards leaving it as is in the options stated by @tessus. Appreciate the feedback, and let's determine the best direction.

tessus:
> Replacing spaces with underscores in filenames might be practical for URLs but moves away from this PR's intent

I do understand what you are saying, but I am trying to explain that when you merge this PR and later on it is decided that replacing spaces with underscores is a good idea anyway, the code that was merged for this PR becomes cruft and useless, since there won't be any spaces anymore in filenames.

This PR showed how useless and annoying spaces are, so why keep them. There is no reason to. As mentioned before, rustypaste is not a storage backend which goal it is to preserve filenames. Filenames were not the goal of this project, neither is the goal to preserve spaces in filenames. This is not why rustypaste was created.

Anyhoo, I don't eant to dismiss the notion that spaces are important, so the 3rd option I mentioned allows for both situations. It didn't mean that one has to replace the spaces with `%20` in filenames, but rather use the codepath currently in this PR.

I would like to add a PR in the future that replaces spaces with underscores. However, if this is done automatically this code becomes obsolete. To not do that another config option is necessary. But why not use the same config option to handle both cases? 

Maybe the name of the option I suggested was off. Maybe it should be 

`paste.filenames` and the 2 options: `encode` (your codepath), `replace_spaces` (my codepath).

Or maybe you can come with a better naming. The current one can only be used for encoding, since the word `enccode` is part of the config option. To allow for both it should be more generic.

Another idea (although not very good, bt it is just as an example):
`paste.algorithm`:  `url_encode`, `replace_spaces`

P.S.: The reason why I think one config option makes sense is that the one exludes the other.

omerbustun:
@tessus,

I've updated the PR to provide three options for handling spaces in filenames: `encode`, `replace`, and `none`. I believe this addresses both of our concerns. Additionally, I've added unit and fixture tests corresponding to the latest changes.

I look forward to seeing it merged.

Best,
Ömer

tessus:
Thanks for the change. I have 2 comments:

We only need 2 options. `none` is irrelevant, because in that case the option is just not set. (commented out)

The possible values for that option should be handled via an enum:

```rust
enum SpaceHandling {
    Encode,
    Replace
}

impl SpaceHandling {
    fn as_str(&self) -> &'static str {
        match self {
            SpaceHandling::Encode => "encode",
            SpaceHandling::Replace => "replace"
        }
    }
}
```

I think this is the Rust way, but orhun would know better.

src/paste.rs
```diff
@@ -233,7 +245,10 @@ impl Paste {
                     .to_string());
             }
         }
-        Ok(self.store_file(file_name, expiry_date, &config)?)
+
+        let file_name = process_filename_for_storage(file_name, &config);
```
tessus, line 249: This is not needed, because this function is called in `store_file`. If you left it here, the same funtion would be called twice.

omerbustun: Thank you for highlighting that. I've made the necessary correction and updated the commit.

tessus:
I believe the `as_str` method is not even necessary, if you use:

```rust
#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
#[serde(rename_all = "lowercase")]
pub enum SpaceHandling {
    Encode,
    Replace,
}
```

There is a similar code block in `random.rs`. But let's wait for orhun's input.

Btw, I am also more than happy to make the "enum" change, if you allow me to push to your branch.

orhun:
Yeah having an `enum` for this would be nice I think.

src/server.rs
```diff
@@ -167,6 +167,16 @@ async fn version(
     Ok(HttpResponse::Ok().body(version.to_owned() + "\n"))
 }
 
+fn process_filename_for_url(file_name: &str, config: &RwLock<Config>) -> String {
```
orhun, line 170: I didn't understand why we have 2 almost identical process functions. Can you use an enum and also refactor this?

fixtures/test-list-files/test file with spaces.txt
```diff
@@ -0,0 +1 @@
+test data
```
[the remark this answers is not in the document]
omerbustun: The `teardown` function in the fixture tests already ensures the removal of the created test files:

```rust
teardown() {
  rm "test file with spaces.txt"
  rm -r upload
}
```

The file was likely included in the commit due to a test interruption on my side, which prevented the `teardown` from executing. I'll remove it in the next commit.

omerbustun:
I've updated the code to use an `enum` and made changes to `server.rs` and `paste.rs` based on your suggestions. I moved the relevant parts to `util.rs`, though I'm not sure if this fits the project's structure. I added new unit tests and adjusted the existing fixture tests. I'm looking forward to your feedback.

Also, since I'm quite new to Rust and this is my first PR, I really appreciate your patience and guidance so far :).

src/util.rs
```diff
@@ -103,6 +103,29 @@ pub fn sha256_digest<R: Read>(input: R) -> Result<String, ActixError> {
...
+/// If `handling` is `Some(SpaceHandling::Encode)`, spaces in the filename are replaced with "%20".
+/// If `handling` is `Some(SpaceHandling::Replace)`, spaces in the filename are replaced with underscores.
+/// If `handling` is `None`, the filename is returned unchanged.
+pub fn process_filename(file_name: &str, handling: Option<SpaceHandling>) -> String {
+    match handling {
+        Some(SpaceHandling::Encode) => file_name.replace(' ', "%20"),
+        Some(SpaceHandling::Replace) => file_name.replace(' ', "_"),
+        None => file_name.to_string(), // keep the filename unchanged
+    }
+}
```
orhun, line 127: You can possibly make this function a part of of the `SpaceHandling` enum.

tessus (commented):
LGTM.

I also support orhun's idea to add the function to the enum. Then this should be good to go.

src/util.rs
omerbustun: Addressed in [94ae863](https://github.com/orhun/rustypaste/commit/94ae863f61b0b59f4ab9959abc828a39960bb0e4).

src/util.rs
```diff
@@ -191,22 +192,16 @@ mod tests {
 
     #[test]
     fn test_filename_replace() {
-        let handling = Some(SpaceHandling::Replace);
-        let processed_filename = process_filename("file with spaces.txt", handling);
-        assert_eq!(processed_filename, "file_with_spaces.txt"); // Check if the space is replaced with underscore
+        let handling = SpaceHandling::Replace;
+        let processed_filename = handling.process_filename("file with spaces.txt");
```
tessus, line 196:
```suggestion
        let processed_filename = SpaceHandling::Replace.process_filename("file with spaces.txt");
```

src/util.rs
```diff
@@ -191,22 +192,16 @@ mod tests {
...
     }
-
+    
     #[test]
     fn test_url_encode() {
-        let handling = Some(SpaceHandling::Encode);
-        let encoded_filename = process_filename("file with spaces.txt", handling);
-        assert!(encoded_filename.contains("%20")); // Check if the space is encoded
+        let handling = SpaceHandling::Encode;
+        let encoded_filename = handling.process_filename("file with spaces.txt");
```
tessus, line 203:
```suggestion
        let encoded_filename = SpaceHandling::Encode.process_filename("file with spaces.txt");
```

tessus (commented):
Other than the 2 inline comments, it looks great. My suggestions only remove 2 unnecessary one line assignments. It's not critical, but I think it's best to avoid those. I made the same mistake myself before. ;-)

Thanks for this PR!

omerbustun:
Thank you for your feedback, @tessus! I've incorporated the suggested changes and made sure everything is aligned by running both clippy and fmt on the final commit. Let me know if there's anything else!

src/util.rs
omerbustun: Addressed in [c716f33](https://github.com/omerbustun/rustypaste/commit/c716f3333b63f84f81e53f5a5609092215c32d56).

src/util.rs
omerbustun: Addressed in [c716f33](https://github.com/omerbustun/rustypaste/commit/c716f3333b63f84f81e53f5a5609092215c32d56).

tessus (commented):
LGTM. I don't see anything else. Unless I missed something this is good to go.

src/server.rs
omerbustun: Addressed in [84626fe](https://github.com/omerbustun/rustypaste/commit/84626fe45187d61b3e2f2ed82cf63ea04ee89660).

orhun (approved):
Thanks for the implementation/discussion @omerbustun @tessus! 💖

PR #109, opened 2023-08-08
fix(server): do not list expired files

tessus:
## Description

Expired files are also returned by the `/list` endpoint. This change only returns valid and accessible files.

## Motivation and Context

Fix above mentioned issue.

## How Has This Been Tested?

- autmated tests
- manual testing

## Changelog Entry

````
### Fixed

- Do not list expired files
````

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -18,6 +18,7 @@ use std::env;
 use std::fs;
 use std::path::PathBuf;
 use std::sync::RwLock;
+use std::time::Duration;
 use uts2ts;
 
 /// Shows the landing page.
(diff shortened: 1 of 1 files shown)


orhun:
Thanks for submitting this PR!

I refactored the code in 67c2de3573eac3e2f491803c7fe7667efff60f49

Do you mind adding some unit tests to cover this change? Then I think we're good to go! 🐻

tessus:
I will try adding a test, but have to read up on sleep. I think I might be able to use https://doc.rust-lang.org/std/thread/fn.sleep_ms.html

However I don't fully understand your refactored code. In the beginning I thought about doing the same, but this only sets `expires_at_utc` to `None`, which is valid when the file does not have an extension (file does not expire). So now the code just list an expired file as a file without expiration.
Unless I am missing something.

orhun:
Hmm, have you tested the changes? Is it really what is happening :o

>  but this only sets expires_at_utc to None

`return None;` is returning `None` to the `and_then` closure above which means the file should be skipped.

tessus:
> Hmm, have you tested the changes? Is it really what is happening :o
> 
> > but this only sets expires_at_utc to None
> 
> `return None;` is returning `None` to the `and_then` closure above which means the file should be skipped.

I was compiling the refactored code while writing the reply. Yes, you are right. The code works.
I didn't know that `return None;` id returning it to the `and_then` closure. I was looking for such a construct, somwthing like `continue` or `skip`, but found nothing. I thought the return would return it to the assignment above. Interesting.

tessus:
> Do you mind adding some unit tests to cover this change? Then I think we're good to go! 🐻

Done.

tessus:
> `return None;` is returning `None` to the `and_then` closure above which means the file should be skipped.

But isn't `return None;` and the `None` on line 347 the same? So does that mean that `None` will also be returned to the closure, if there is no extension (no expiry for a file)?

If this is a case, this means files without an expiry date/time will never be listed.

tessus:
Nope, I have tested it. They are not the same, which is confusing. I guess I still have a long way to go with Rust.

orhun:
Yeah, it looks a bit weird at first glance but it works 🐻

orhun (approved):

PR #112, opened 2023-08-13
fix(server): don't log invalid token in release builds

### LINKED ISSUE #111: server logs token, when wrong token is sent
I have noticed that the server logs a sent token, if the token is not valid:

e.g.: `rpaste -a invalid_token -V`

```
[2023-08-13T19:24:30Z WARN  rustypaste::auth] authorization failure for a.b.c.d (header: invalid_token)
```

There are 2 issues with this:

- the text is wrong. `header` should print the entire header or the text should be renamed to `token`
- the token should not be printed at all

I don't think a token should be logged, even if it is an invalid one. At least not in a release build. Such a situation can arise when you use rpaste with the token in the config file, but use different rustypaste servers. If you forget only once to add the -a flag, a production token for another instance will be logged.

I suggest to remove everything after the IP address.

However, I do think that it could be important for testing/debugging purposes so let's just add the token for failures in debug builds.

What do you think?

tessus:
## Description

The server logs a sent token, if the token is not valid:

e.g.: `rpaste -a invalid_token -V`

```
[2023-08-13T19:24:30Z WARN  rustypaste::auth] authorization failure for a.b.c.d (header: invalid_token)
```

This change prints the token only in debug mode. In release mode the log entry will look like this:

```
[2023-08-13T19:24:30Z WARN  rustypaste::auth] authorization failure for a.b.c.d
```

## Motivation and Context

fixes #111 

Let's say you use rpaste with the token in the config file (or curl with the auth_token as env var or in a file), but use different rustypaste servers. If you forget only once to add the `-a` flag (or add the header to curl), a production token for another instance will be logged.

## How Has This Been Tested?

- cargo test
- fixtures
- manual testing

## Changelog Entry

````
### Fixed

- Don't log invalid token in release builds
````

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -11,11 +11,14 @@ pub fn check(host: &str, headers: &HeaderMap, tokens: Option<Vec<String>>) -> Re
             .map(|v| v.to_str().unwrap_or_default())
             .map(|v| v.split_whitespace().last().unwrap_or_default());
         if !tokens.iter().any(|v| v == auth_header.unwrap_or_default()) {
+            #[cfg(debug_assertions)]
             log::warn!(
-                "authorization failure for {} (header: {})",
+                "authorization failure for {} (token: {})",
                 host,
                 auth_header.unwrap_or("none"),
             );
+            #[cfg(not(debug_assertions))]
+            log::warn!("authorization failure for {}", host);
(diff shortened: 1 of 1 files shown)


orhun (approved):
Neat 💯

PR #136, opened 2023-08-31
feat(server): add delete endpoint

### LINKED ISSUE #127: DELETE end point
I'm using rustypaste on my LAN as a lightweight file server for transferring files/links/text across devices (currently I'm building the UI for it). One thing I realized rustypaste lacks which I would like to have is the ability to remove files and the ability to update files (change their contents or set a new expiration).

However, I'm aware that my use case may be niche and probably not something that would be included in most pastebin services, so before I set about building a PR for this I thought I'd see if this fits with the vision of the app enough to consider. If not I may just need to find a new backend service for my UI.

Interestingly I have found that there's not a lot out there in the ways of a simple file hosting service that have simple API access without a lot of other bells and whistles that I don't need. Most solutions are pretty heavy (like syncthing) which is why I ended up looking at pastebin services. Rustypaste fits quite nice in this regard with it's simple HTTP interface and speed. 

EDIT: thinking more on this, I think the only one I would truly need is DELETE.

ajbdev:
## Description

Add a DELETE endpoint guarded by a separate auth token array.

## Motivation and Context

Closes #127 

## How Has This Been Tested?

Wrote a test case and tested manually.

## Changelog Entry

````
### Added

- Change `Config.get_tokens` to `Config.get_auth_tokens` and update invocations.
- Add delete_tokens array to config and `get_delete_tokens` method to Config for retrieving delete tokens.
- Add DELETE endpoint that will remove files if a delete token is set and authorized by the request.
- Test for new DELETE endpoint
- Add DELETE_TOKEN_ENV for setting the delete tokens array from an ENV var.
````
-->

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,7 @@ Here you can read the blog post about how it is deployed on Shuttle: [https://bl
     - supports overriding and blacklisting
     - supports forcing to download via `?download=true`
   - no duplicate uploads (optional)
-  - listing files
-  - deleting files
+  - listing/deleting files
   - custom landing page
 - Single binary
   - [binary releases](https://github.com/orhun/rustypaste/releases)
--- a/README.md
+++ b/README.md
@@ -50,8 +50,9 @@ Here you can read the blog post about how it is deployed on Shuttle: [https://bl
     - [URL shortening](#url-shortening)
     - [Paste file from remote URL](#paste-file-from-remote-url)
     - [Cleaning up expired files](#cleaning-up-expired-files)
+    - [Delete file from server](#delete-file-from-server)
   - [Server](#server)
-  - [List endpoint](#list-endpoint)
+    - [List endpoint](#list-endpoint)
     - [HTML Form](#html-form)
     - [Docker](#docker)
     - [Nginx](#nginx)
@@ -79,6 +80,7 @@ Here you can read the blog post about how it is deployed on Shuttle: [https://bl
     - supports forcing to download via `?download=true`
   - no duplicate uploads (optional)
   - listing files
+  - deleting files
   - custom landing page
 - Single binary
   - [binary releases](https://github.com/orhun/rustypaste/releases)
@@ -235,6 +237,14 @@ while read -r filename; do
 done
 ```
 
+#### Delete file from server
+
+Set `delete_tokens` array in [config.toml](./config.toml) to activate the DELETE endpoint and secure it with one (or more) auth token(s).
+
+```sh
+$ curl -H "Authorization: <auth_token>" -X DELETE "<server_address>/file.txt"
+```
+
 ### Server
 
 To start the server:
@@ -260,7 +270,7 @@ You can also set multiple auth tokens via the array field `[server].auth_tokens`
 
 See [config.toml](./config.toml) for configuration options.
 
-### List endpoint
+#### List endpoint
 
 Set `expose_list` to true in [config.toml](./config.toml) to be able to retrieve a JSON formatted list of files in your uploads directory. This will not include oneshot files, oneshot URLs, or URLs.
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by tessus — COMMENTED** [review]
LGTM

closes #127

**Comment by tessus** [comment]
I was thinking a bit more. Maybe we should create an enum for token_type: auth, delete. And then use just one function `get_tokens(&self, enum)`. I have no idea, whether this is better code or me just trying to optimize without any gain.

For me thia PR is perfecctly fine. Let's see what @orhun has to say.

File: README.md [hunk 1]
```diff
@@ -79,6 +80,7 @@ Here you can read the blog post about how it is deployed on Shuttle: [https://bl
     - supports forcing to download via `?download=true`
   - no duplicate uploads (optional)
   - listing files
+  - deleting files
```
orhun, line 83:
```suggestion
  - listing/deleting files
```

README.md
```diff
@@ -235,6 +237,14 @@ while read -r filename; do
 done
 ```
 
+#### Delete file from server
+
+Set `delete_tokens` array in [config.toml](./config.toml) to activate the DELETE endpoint and secure it with one (or more) auth token(s).
```
orhun, line 242:
```suggestion
Set `delete_tokens` array in [config.toml](./config.toml) to activate the [`DELETE`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE) endpoint and secure it with one (or more) auth token(s).
```

src/server.rs
```diff
@@ -146,6 +146,33 @@ async fn serve(
...
+    let tokens = config.get_delete_tokens();
+    if tokens.is_none() {
+        log::warn!("delete endpoint not served because there are no delete_tokens set");
+        return Err(error::ErrorForbidden("endpoint is not exposed\n"));
+    }
+    auth::check(host, request.headers(), tokens)?;
+    if !path.is_file() || !path.exists() {
+        return Err(error::ErrorNotFound("file is not found or expired :(\n"));
+    }
+    fs::remove_file(path).expect("Could not delete file");
```
orhun, line 172: Can we use `error!` here to log the error and return 500?

src/server.rs
```diff
@@ -146,6 +146,33 @@ async fn serve(
...
+    let config = config
+        .read()
+        .map_err(|_| error::ErrorInternalServerError("cannot acquire config"))?;
+    let path = config.server.upload_path.join(&*file);
+    let path = util::glob_match_file(path)?;
+    let connection = request.connection_info().clone();
+    let host = connection.realip_remote_addr().unwrap_or("unknown host");
+    let tokens = config.get_delete_tokens();
+    if tokens.is_none() {
+        log::warn!("delete endpoint not served because there are no delete_tokens set");
```
orhun, line 165: Are we testing this case with unit tests/fixtures? (not strictly needed though)

orhun (commented):
Overall looks good to me. Just had a few comments 🐻 

@tessus I don't know what would be the benefit of adding an `enum` but we can go with that for the sake of having a single function for retrieving the tokens. @ajbdev can you also make that change?

tessus:
> @tessus I don't know what would be the benefit of adding an enum but we can go with that for the sake of having a single function for retrieving the tokens. @ajbdev can you also make that change?

I came up with this "brilliant" idea, thus I'll take care of it. ;-)

src/server.rs
tessus: done

src/server.rs
tessus: done

tessus:
@orhun all done!

tessus:
@orhun It seems that gh had a connection issue during 2 tests. Any chance you can rerun the action?

If I run the tests locally, everything is fine.

orhun (approved):
Tested and works well!

Thank you both @ajbdev @tessus 🚀

PR #199, opened 2023-12-08
refactor!: cleanup authorization boilerplate

DDtKey:
## Description

- fix not-exposed endpoints to return `404` instead of `403` (they literally not intended to be exposed) - **the only breaking change, can be extracted to separate PR**
- keep authz rules cohesive 
- clean up boilerplate code using macro from `actix-web-grants` to protect endpoints
- utilize `ErrorHandlers` to log unauthorized errors instead of in-place `host` handling & etc

## Motivation and Context

I came across `rustypaste` and decided to check the code a little & play locally with it. As a result, some of the authorization rules were not obvious to me. For example:
- lack of `delete_tokens` - hides the endpoint
- lack of `auth_tokens` - allows the access (btw, is it intended? because sound like a bug)
- disabled endpoints return 403, but from the point of view of security and correctness it is more likely to be 404 - we literally do not expose them

So I decided to do a little refactoring, which I hope you find useful.

Another part of the changes is the introduction of [actix-web-grants](https://crates.io/crates/actix-web-grants) - this is a small but quite flexible library, with which it will also be easy to change the rules, for example, switch to RBAC/ABAC or something else (actually I'm maintainer of the library, so it's may be subjective)

## How Has This Been Tested?

Mostly by current test-coverage, I kept the same behavior, except `403` -> `404`

## Types of Changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other 

## Checklist:


orhun:
Definitely sounds/looks useful! I can review the code tomorrow.

One thing to note here, you can [add fixture tests](https://github.com/orhun/rustypaste/tree/master/fixtures) to test the new behavior.

DDtKey:
Thanks for the reference!

 I took a look at fixtures and think it makes sense to add new fixtures  for not-exposed endpoints (`list`, `version` or `delete-tokens` not configured)

Otherwise I have mostly kept the behavior and there is coverage of the main cases.

Except probably one case, which hasn’t been covered before: if `auth-tokens` unset/not configured - access is allowed (just to clarify, is this intended?)

There are unit tests ofc, but I’ll add fixtures a bit later!

src/config.rs
```diff
@@ -148,29 +150,37 @@ impl Config {
...
+    pub fn get_tokens(&self, token_type: TokenType) -> Option<HashSet<String>> {
+        let tokens = match token_type {
             TokenType::Auth => {
-                let mut tokens = self.server.auth_tokens.clone().unwrap_or_default();
+                let mut tokens: HashSet<_> = self.server.auth_tokens.clone().unwrap_or_default();
+
                 #[allow(deprecated)]
                 if let Some(token) = &self.server.auth_token {
-                    tokens.insert(0, token.to_string());
+                    tokens.insert(token.to_string());
```
orhun, line 160: Is it possible to insert as the first element as before?

src/config.rs
```diff
@@ -148,29 +150,37 @@ impl Config {
...
+                    }
+                    _ => {}
                 }
                 tokens
             }
         };
-        tokens.retain(|v| !v.is_empty());
-        (!tokens.is_empty()).then_some(tokens)
+
+        Some(tokens).filter(|v| !v.is_empty())
```
orhun, line 183: Can you add an unit test to verify this behavior?

src/server.rs
```diff
@@ -181,27 +176,23 @@ async fn delete(
...
         .read()
         .map_err(|_| error::ErrorInternalServerError("cannot acquire config"))?;
-    let connection = request.connection_info().clone();
-    let host = connection.realip_remote_addr().unwrap_or("unknown host");
-    let tokens = config.get_tokens(TokenType::Auth);
-    auth::check(host, request.headers(), tokens)?;
     if !config.server.expose_version.unwrap_or(false) {
         warn!("server is not configured to expose version endpoint");
-        Err(error::ErrorForbidden("endpoint is not exposed\n"))?;
+        Err(error::ErrorNotFound("endpoint is not exposed\n"))?;
```
orhun, line 186: If we don't want to expose anything, should we remove the message as well and just return 404?

src/auth.rs
```diff
@@ -1,55 +1,162 @@
...
+            warn!("delete endpoints is not served because there are no delete_tokens set");
+            return Err(error::ErrorNotFound("endpoint is not exposed\n"));
         }
     }
-    Ok(())
+
+    Ok(user_tokens)
+}
+
+pub(crate) fn unauthorized_error() -> actix_web::HttpResponse {
```
orhun, line 48: nit: document this function.

orhun:
> fix not-exposed endpoints to return 404 instead of 403 (they literally not intended to be exposed) - the only breaking change, can be extracted to separate PR

> disabled endpoints return 403, but from the point of view of security and correctness it is more likely to be 404 - we literally do not expose them

I think it makes sense. We can mention this in the changelog as breaking.

> lack of delete_tokens - hides the endpoint

The intention was to require tokens to delete files on the server. Do you think we should change that?

> lack of auth_tokens - allows the access (btw, is it intended? because sound like a bug)
> Except probably one more case, which hasn’t been covered before: if auth-tokens unset/not configured - access is allowed (just to clarify, is this intended?)

Yeah, this is intended. Do you think we should have an empty list of tokens for allowing access? It is not a big deal imo.

> Another part of the changes is the introduction of [actix-web-grants](https://crates.io/crates/actix-web-grants)

I saw a derive macro on the endpoints that we want to protect but can you tell me how that works? Also it would be nice to get some information in terms of which version of Actix do you support, what are the plans for the future (will there be any breaking changes) etc.

> I took a look at fixtures and think it makes sense to add new fixtures for not-exposed endpoints (list, version or delete-tokens not configured)
> There are unit tests ofc, but I’ll add fixtures a bit later for all these cases!

Great! Just let me know if you have problems / need guidance while adding them.

src/config.rs
DDtKey: Btw, what was the point of this? 

I thought it was done this way in order to optimize the access to the vector, but with `HashSet` we have `O(1)` and don't have duplicates 🤔

src/server.rs
DDtKey: It makes a lot of sense to me

DDtKey:
> > lack of delete_tokens - hides the endpoint
> 
> The intention was to require tokens to delete files on the server. Do you think we should change that?

Not sure, the main confusion is that we have `expose` flag for `list` & `version` - which is really clear, but in case of `delete` it depends on `delete_tokens`. Not so clear when configuring the app, the first thing that comes to mind is that it should be `401` (any token won't match configured ones, they are unset), but in fact `404` seems safer in this case.

So I think we can keep it as is, probably improve the documentation related to the `config` - if `delete_tokens` unset then `delete` endpoint won't be exposed at all.


> > lack of auth_tokens - allows the access (btw, is it intended? because sound like a bug)
>
> Yeah, this is intended. Do you think we should have an empty list of tokens for allowing access? It is not a big deal imo.

I thought more about explicit configuration flag, like `disable_auth`. Just looks less fragile when configuring the app, but the behavior itself is ok, just there was no any comments and I have been concerned. I've added comment & logging with this behavior anyway. 


> I saw a derive macro on the endpoints that we want to protect but can you tell me how that works? 

Actually it works pretty straightforward:
- Middleware which extracts some "authority" entity from request (in our case we do extract `TokenType`)
- it adds specific type (`AuthDetails<T>`) to extensions 
- macro actually helps to get rid of the following boilerplate, it expands to something like that:
```rust
async fn endpoint(details: AuthDetails) -> HttpResponse {
    if details.has_authority(TokenType::Auth) {
        // body of our function
    } else {
       // error we want to see, defaults to `Forbidden`
    }
}
```
But we can write it manually as well 

> Also it would be nice to get some information in terms of which version of Actix do you support, what are the plans for the future (will there be any breaking changes) etc.

`actix-web` was the first  framework within this ecosystem, there is mention here: https://github.com/DDtKey/protect-endpoints/tree/main/actix-web-grants#supported-actix-web-versions

Recently I've delivered a huge update and I'd say it's likely long-term vision of API. As part of that update I also switched to `release-plz` (with `cargo-semver-checks` & `git-cliff` under the hood), so we have the validations.

Current macro-syntax flexible enough and should fit a lot of cases, including ABAC & RBAC, for example, you can combine the conditions in Rust-friendly way, using `all` & `any`:

``` rust
// snippet from examples
#[actix_web_grants::protect(any("ADMIN", expr = "user.is_super_user()"), ty = "Role")]
async fn admin_or_super_user(user_id: web::Path<i32>, user: web::Data<User>) -> HttpResponse {
    HttpResponse::Ok().body("some secured response")
}
```

So **I don't expect the breaking changes in the near future**, and if they would be - it will be major (semver) update without breaking existing applications, and with back-porting the changes if it's compatible & makes sense.

src/config.rs
DDtKey: Sure!
Btw, we do filter out only empty string (`""` ), don't we want to filter whitespace strings and etc? 🤔

DDtKey: I changed to filter blank strings as well, it looks more secure to me

src/auth.rs
DDtKey: Added

DDtKey:
> Great! Just let me know if you have problems / need guidance while adding them.

I've added fixtures, it looks pretty clear. 

Some not related observations:
- Would be nice to call `teardown` in case of failure, to clean up
- macos: `sleep` doesn't work with time-units (e.g `sleep 1s`)
- macos: there is no `fallocate`  (alternative would be to use `dd if=/dev/random of=normalfile count=10000 bs=1024 status=none`)

I think would be nice to ensure it works on any platform, or provide fast way to run it with docker for example

src/config.rs
orhun: It was just to prioritize checking the tokens, so if auth_token is given via config then it will be checked first etc.

I think this is something trivial though, feel free to ignore.

src/config.rs
orhun: Yeah, I think not allowing whitespace strings makes perfect sense. Thanks for adding the tests!

orhun:
> So I think we can keep it as is, probably improve the documentation related to the config - if delete_tokens unset then delete endpoint won't be exposed at all.

Makes sense - would you be interested in making a PR to update README.md about this?

> I thought more about explicit configuration flag, like disable_auth [...] I've added comment & logging with this behavior anyway.

👍🏼 

> Actually it works pretty straightforward:

Thanks for the explanation! It is more clear now.

> I also switched to release-plz (with cargo-semver-checks & git-cliff under the hood)

Aye [that's](https://github.com/orhun/git-cliff) me! Let me know if you need help with any configuration.

> Current macro-syntax flexible enough and should fit a lot of cases, including ABAC & RBAC, for example, you can combine the conditions in Rust-friendly way, using all & any:

Wow that's nice!

> So I don't expect the breaking changes in the near future, and if they would be - it will be major (semver) update without breaking existing applications, and with back-porting the changes if they're compatible & makes sense.

Alright, sounds good.

src/config.rs
DDtKey: Yeah, I think it's not applicable anymore, we do use `hashset.contains(..)` now, instead of `.iter().any(..)`, so the order doesn't matter actually anymore. Thus, we directly check if token in `Authorization` headers matches any of configured ones with `O(1)` complexity

DDtKey:
> Makes sense - would you be interested in making a PR to update README.md about this?

Take a look at #202, please

> Aye [that's](https://github.com/orhun/git-cliff) me! Let me know if you need help with any configuration.

Yes, thanks for the wonderful tool! :pray:  Actually I've been using it for about a year on different projects!  😅

orhun:
LGTM! Just one last thing.

I usually put the user-facing changes in the changelog so I guess the only things that have changed are:

- we now return 404 instead of 403 for non-exposed endpoints
- we don't allow blank (whitespace) tokens 

Do we need to mention something else in the changelog?

DDtKey:
In that case I think these two are enough, I'll update the changelog entr

orhun (approved):
Thanks!

PR #200, opened 2023-12-09
test(fixtures): support `fixtures` on `macOS`

DDtKey:
## Description

Change fixtures to support macos as well as linux

## Motivation and Context

- `sleep` doesn't work with time-units (e.g `sleep 1s`)
- there is no `fallocate`  (`dd if=/dev/random of=normalfile count=10000 bs=1024 status=none` used)
- string formatting with substitution works incorrectly `"$(curl -s -H "Authorization: $auth_token" localhost:8000/list | grep -o 'file_name' | wc -l)"` returns `"     3"`, so it's not equal to `"3"`

## How Has This Been Tested?

Fixtures works on macOS and should pass CI (on lunux)

## Changelog Entry

````
### Changed

- Support fixtures on macOS
````

## Types of Changes

- [X] Other  - Testing framework

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [X] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/fixtures/test-server-payload-limit/test.sh
+++ b/fixtures/test-server-payload-limit/test.sh
@@ -3,7 +3,8 @@
 setup() {
   touch emptyfile
   truncate -s 9KB smallfile
-  fallocate -l 10000 normalfile
+#  fallocate -l 10000 normalfile
+  dd if=/dev/random of=normalfile count=10000 bs=1024 status=none
   truncate -s 11KB bigfile
 }
 
--- a/fixtures/test-server-payload-limit/test.sh
+++ b/fixtures/test-server-payload-limit/test.sh
@@ -3,7 +3,7 @@
 setup() {
   touch emptyfile
   truncate -s 9KB smallfile
-#  fallocate -l 10000 normalfile
+  # On Linux, `fallocate -l 10000 normalfile` can be used for a better precision.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by DDtKey** [comment]
Though it can be broken anytime without CI, so if it makes sense I can add a matrix with OS

**Comment by orhun** [comment]
Yeah, I think it is a good idea to add a matrix here:

https://github.com/orhun/rustypaste/blob/3e342068840ba2ba5857deb9e4bacf80389dae65/.github/workflows/ci.yml#L47

File: fixtures/test-server-payload-limit/test.sh [hunk 1]
```diff
@@ -3,7 +3,8 @@
 setup() {
   touch emptyfile
   truncate -s 9KB smallfile
-  fallocate -l 10000 normalfile
+#  fallocate -l 10000 normalfile
```
orhun, line 6:
```suggestion
  # On Linux, `fallocate -l 10000 normalfile` can be used for a better precision.
```

orhun (commented):
Let's just add a comment there 🐻

orhun (approved):
Thanks for your contribution!

PR #234, opened 2024-01-30
fix(server): return the correct file on multiple files with same name

### LINKED ISSUE #218: file not found, even though on server and not expired
# Steps to reproduce

1. set `delete_expired_files = { enabled = true, interval = "1h" }` in `config.toml`
2. no random filenames in `config.toml`
3. start rustypaste
4. `echo "test content" >test.txt`
5. `curl -F "file=@test.txt" -H "expire:1s" http://127.0.0.1:8000`
6. `curl -F "file=@test.txt" -H "expire:1s" http://127.0.0.1:8000`
7. `curl -F "file=@test.txt" -H "expire:1min" http://127.0.0.1:8000`
8. `curl http://127.0.0.1:8000/test.txt`

# Expected

```
test content
```

# What happened

```
file is not found or expired :(
```

However, the file is shown when using the list endpoint with the correct expiry date (1 min in the future).

tessus:
## Description

The `glob_match_file` function returns the wrong entry. `glob` yields paths in alphabetical order. We want the file with the longest expiry date/time.

## Motivation and Context

fixes #218 

## How Has This Been Tested?

- manually
- cargo test
- fixtures

## Changelog Entry

````
### Changed

- Fixed an erroneous `file not found` occurrence
````

## Types of Changes

- [ ] Other 

## Checklist:


fixtures/test-fix-218/config.toml
orhun, line 1: Can we rename this to something more explanatory? `test-fix-218` might not make sense if someone else takes a look.

orhun (commented):
Such an effective fix! Can you think of any edge cases that this would break backwards compatibility?

fixtures/test-fix-218/config.toml
tessus: I just didn't come up with a proper name. I'll try to think harder. ;-)

tessus:
I thought about it for a while and it won't break anything.

But I see an issue that has nothing to do with this or the previous PRs: One can upload multiple files with the same name, but that was already possible with the `curl -F "file=@x.txt;filename=wow.nice"` command when `random_url` was not set. This case was never taken into consideration.
We can look into this in a separate discussion or issue.

fixtures/test-fix-218/config.toml
tessus: the new name is long, but I couldn't think of anything else

tessus:
I created a separate discussion #241 

> Can you think of any edge cases that this would break backwards compatibility?

A bit mor info:

If multiple files with the same name are on the server, the file with the longest expiry timestamp will be returned when accessing the file via `curl <server>/file.txt`

Previously a different file was returned. Which one is not entirely clear depending on expiration but not yet cleaned from disk, and so on.

But this was never a defined behavior anyway, thus no breaking change.

orhun (approved):
Thanks for looking into this and creating a discussion! I will get to that soon - I think this PR is in good shape to merge now.

PR #210, opened 2023-12-20
fix(auth): remove excessive warning and typo

tessus:
## Description

Iff one doesn't use auth tokens, their log is spammed with warning messages.
This change removes the warning message.

It also fixes a small typo.

## Motivation and Context

Fixes #209

## How Has This Been Tested?

cargo test
fixtures

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (no code change)
- [ ] Refactor (refactoring production code)
- [ ] Other 

## Checklist:

- [x] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly.
- [x] I have formatted the code with [rustfmt](https://github.com/rust-lang/rustfmt).
- [x] I checked the lints with [clippy](https://github.com/rust-lang/rust-clippy).
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #236, opened 2024-01-31
ci(shuttle): fix deployment

tessus:
## Description

This change fixes the attempt to deploy on a PR.

## Motivation and Context

see https://github.com/orhun/rustypaste/pull/210#issuecomment-1864888922

## How Has This Been Tested?

on my own repo

## Changelog Entry

````
### Changed

- Do not run shuttle deployment on Pull Requests
````

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/.github/workflows/shuttle.yml
+++ b/.github/workflows/shuttle.yml
@@ -21,12 +21,6 @@ jobs:
           toolchain: stable
           profile: minimal
           override: true
-      - uses: dorny/paths-filter@v3
-        id: changes
-        with:
-          filters: |
-            manifest:
-              - 'Cargo.**'
       - name: Install cargo-binstall
         uses: taiki-e/install-action@cargo-binstall
       - name: Install cargo-shuttle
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tessus** [comment]
of course there is an unrelated (and false, due to network issues) error in another pipeline.

please re-run this workflow.

File: .github/workflows/shuttle.yml [hunk 1]
```diff
@@ -40,11 +37,7 @@ jobs:
       - name: Build
         run: cargo build --locked --verbose
       - name: Deploy
-        if: |
```
orhun, line 43: If we are removing this check then it would make sense to remove the `dorny/paths-filter` step as well. The idea was that I only want to run the shuttle deployments when there is a change made to `Cargo.toml`.

tessus: ok, then I'll remove it, unless you can think of another way.

tessus: done

orhun (approved):
Thanks!

PR #247, opened 2024-02-16
feat(server): do path joins more safely

RealOrangeOne:
## Description

Check that a joined path is still a child of the original path.

## Motivation and Context

When joining paths, there's little protection against directory traversal. From testing, Actix seems to nicely sanitises these in the request, which is great, but it's still better to add a little protection.

## How Has This Been Tested?

Unit tests have been added, which pass.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -71,7 +71,11 @@ fn setup(config_folder: &Path) -> IoResult<(Data<RwLock<Config>>, ServerConfig, 
     // Create necessary directories.
     fs::create_dir_all(&server_config.upload_path)?;
     for paste_type in &[PasteType::Url, PasteType::Oneshot, PasteType::OneshotUrl] {
-        fs::create_dir_all(paste_type.get_path(&server_config.upload_path))?;
+        fs::create_dir_all(
+            paste_type
+                .get_path(&server_config.upload_path)
+                .expect("failed to get path for paste type"),
+        )?;
     }
 
     // Set up a watcher for the configuration file changes.
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,7 +13,7 @@ use rustypaste::util;
 use rustypaste::CONFIG_ENV;
 use std::env;
 use std::fs;
-use std::io::Result as IoResult;
+use std::io::{Error as IoError, ErrorKind as IoErrorKind, Result as IoResult};
 use std::path::{Path, PathBuf};
 use std::sync::{mpsc, RwLock};
 use std::thread;
@@ -71,11 +71,13 @@ fn setup(config_folder: &Path) -> IoResult<(Data<RwLock<Config>>, ServerConfig, 
     // Create necessary directories.
     fs::create_dir_all(&server_config.upload_path)?;
     for paste_type in &[PasteType::Url, PasteType::Oneshot, PasteType::OneshotUrl] {
-        fs::create_dir_all(
-            paste_type
-                .get_path(&server_config.upload_path)
-                .expect("failed to get path for paste type"),
-        )?;
+        let _ = paste_type
+            .get_path(&server_config.upload_path)
+            .ok_or(IoError::new(
+                IoErrorKind::Other,
+                String::from("Invalid upload path"),
+            ))
+            .map(fs::create_dir_all)?;
     }
 
     // Set up a watcher for the configuration file changes.
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,7 +13,7 @@ use rustypaste::util;
 use rustypaste::CONFIG_ENV;
 use std::env;
 use std::fs;
-use std::io::{Error as IoError, ErrorKind as IoErrorKind, Result as IoResult};
+use std::io::Result as IoResult;
 use std::path::{Path, PathBuf};
 use std::sync::{mpsc, RwLock};
 use std::thread;
@@ -71,13 +71,7 @@ fn setup(config_folder: &Path) -> IoResult<(Data<RwLock<Config>>, ServerConfig, 
     // Create necessary directories.
     fs::create_dir_all(&server_config.upload_path)?;
     for paste_type in &[PasteType::Url, PasteType::Oneshot, PasteType::OneshotUrl] {
-        let upload_path = paste_type
-            .get_path(&server_config.upload_path)
-            .ok_or(IoError::new(
-                IoErrorKind::Other,
-                String::from("Invalid upload path"),
-            ))?;
-        fs::create_dir_all(upload_path)?;
+        fs::create_dir_all(paste_type.get_path(&server_config.upload_path)?)?;
     }
 
     // Set up a watcher for the configuration file changes.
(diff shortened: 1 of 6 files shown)


src/util.rs
```diff
@@ -170,4 +185,18 @@ mod tests {
...
+        assert!(safe_path_join("/foo", "bar").is_some());
+        assert_eq!(
+            safe_path_join("/foo/bar", "baz/"),
+            Some("/foo/bar/baz/".into())
+        );
+
+        assert!(safe_path_join("/foo", "/foobar").is_none());
+        assert!(safe_path_join("/foo", "/bar").is_none());
+        assert!(safe_path_join("/foo/bar", "..").is_none());
+        assert!(safe_path_join("/foo/bar", "../").is_none());
```
orhun, line 200: Can you extend the tests for `/foo/../../` + `bar` and different variants of that?

src/util.rs
```diff
@@ -170,4 +185,18 @@ mod tests {
         assert_eq!(Vec::<PathBuf>::new(), get_expired_files(&current_dir));
         Ok(())
     }
+
+    #[test]
+    fn test_safe_join_path() {
```
orhun, line 190: Can you add another unit test for confirming that this util function and standard `join` has the same behavior?

src/paste.rs
```diff
@@ -261,10 +263,12 @@ impl Paste {
...
         }
-        let mut path = self
-            .type_
-            .get_path(&config.server.upload_path)
-            .join(&file_name);
+        let mut path =
+            util::safe_path_join(self.type_.get_path(&config.server.upload_path), &file_name)
+                .ok_or(IoError::new(
+                    IoErrorKind::Other,
+                    String::from("Invalid filename"),
```
orhun, line 270: For consistency:
```suggestion
                    String::from("invalid filename"),
```

src/paste.rs
```diff
@@ -63,7 +63,7 @@ impl PasteType {
         if dir.is_empty() {
             path.to_path_buf()
         } else {
-            path.join(dir)
+            util::safe_path_join(path, Path::new(&dir)).unwrap()
```
orhun, line 66: Let's avoid unwrap :3

orhun (changes requested):

tessus:
@orhun it seems that you have to approve the CI for the 2 PRs.

tessus:
But what I could ascertain from a quick look at the code is that prior to this PR errors were handled gracefully. After this PR rustypaste will panic. Not sure I like this.

orhun:
Yup, I also think that we should handle errors.

RealOrangeOne:
There's only 1 panic this introduces, and it's during server startup. Sure, I could facade it to convert a path issue to `IoResult`, and let Rust deal with returning that one, or I can call `expect`. The latter felt fine for a quick change.

tessus:
What about the occurrences of `PasteType::Url.get_path(&config.server.upload_path).unwrap()`?

Isn't it true that these can panic?

RealOrangeOne:
There are other uses of `unwrap`, but those are strictly in tests, which I've not considered to be an issue. If a test panics, it's functionally the same as a failure, which is what we want.

Yes, the `fs::create_dir_all` is the only new place the application could panic.

tessus:
> but those are strictly in tests

Oops, once again the gh diff failed me again. all the test headers were not in the code, thus I didn't see that those were tests. All good then.

> Yes, the `fs::create_dir_all` is the only new place the application could panic.

I would rather see an error message than a panic. But it's up to orhun.

orhun:
Yes, it would be better to print out an error message than panicking. Also, there are some lints about the usage of `unwrap` in tests 🙂 we can simply replace them with `expect`.

src/main.rs
```diff
@@ -71,7 +71,11 @@ fn setup(config_folder: &Path) -> IoResult<(Data<RwLock<Config>>, ServerConfig,
     // Create necessary directories.
     fs::create_dir_all(&server_config.upload_path)?;
     for paste_type in &[PasteType::Url, PasteType::Oneshot, PasteType::OneshotUrl] {
-        fs::create_dir_all(paste_type.get_path(&server_config.upload_path))?;
+        fs::create_dir_all(
+            paste_type
+                .get_path(&server_config.upload_path)
+                .expect("failed to get path for paste type"),
+        )?;
```
tessus, line 78: There is still a possible panic here.

Sorry for all my comments, but your code introduces a possible panic, while the "old" code just didn't create the dir. Maybe I am missing panics in the "old" code along the path, e.g. when rustypaste tries to save a file when the dir does not exist, but I think it just returns an error message.

RealOrangeOne: There is - yes. I missed this when I wrote the last commit. I'll map it to a sensible error type so it won't panic

src/main.rs
```diff
@@ -71,7 +71,13 @@ fn setup(config_folder: &Path) -> IoResult<(Data<RwLock<Config>>, ServerConfig,
...
     fs::create_dir_all(&server_config.upload_path)?;
     for paste_type in &[PasteType::Url, PasteType::Oneshot, PasteType::OneshotUrl] {
-        fs::create_dir_all(paste_type.get_path(&server_config.upload_path))?;
+        let _ = paste_type
+            .get_path(&server_config.upload_path)
+            .ok_or(IoError::new(
+                IoErrorKind::Other,
+                String::from("Invalid upload path"),
+            ))
+            .map(fs::create_dir_all)?;
```
orhun, line 80: I don't think the error from `create_dir_all` is handled here. I simulated some error with the following snippet:

```rust
let _ = paste_type
    .get_path(&server_config.upload_path)
    .ok_or(IoError::new(
        IoErrorKind::Other,
        String::from("Invalid upload path"),
    ))
    .map(|v| {
        let result = fs::create_dir_all(v);
        IoError::new(IoErrorKind::Other, String::from("some error happened"))
    })?;
```

And didn't see any errors in the application.

Instead, you can probably do this:

```rs
let upload_path = paste_type
    .get_path(&server_config.upload_path)
    .ok_or(IoError::new(
        IoErrorKind::Other,
        String::from("Invalid upload path"),
    ))?;
fs::create_dir_all(upload_path)?;
```

src/paste.rs
```diff
@@ -261,10 +271,19 @@ impl Paste {
...
-            .type_
-            .get_path(&config.server.upload_path)
-            .join(&file_name);
+        let mut path = util::safe_path_join(
+            self.type_
+                .get_path(&config.server.upload_path)
+                .ok_or(IoError::new(
+                    IoErrorKind::Other,
+                    String::from("invalid filename"),
+                ))?,
```
orhun, line 280: I'm thinking if it makes more sense to return `Result<PathBuf>` from `safe_path_join` so that we don't need to map the `Option` everytime we use it. What do you think?

orhun (changes requested):

src/paste.rs
RealOrangeOne: We could, and I guess create our own `IoError` type to handle it. It ought to make the code a bit clearer - I'll give it a try.

src/main.rs
RealOrangeOne: Yep, you're absolutely right. It turns out `and` is what I want. But it's much easier to use 2 separate statements.

orhun:
Looks like there is a test failure:

```
 ---- util::tests::test_safe_join_path stdout ----
thread 'util::tests::test_safe_join_path' panicked at src/util.rs:211:9:
assertion failed: safe_path_join("/foo", "/foobar").is_ok()
stack backtrace:
```

orhun (approved):
Thanks for your contribution!

PR #258, opened 2024-03-08
fix(upload): error on upload with the same filename

tessus:
## Description

Currently it is possible to upload files with the same name. This is undefined behavior. e.g. which file should be returned when accessing it?

To solve this problem this change introduces a check whether a file (that hasn't expired) alredy exists with the same name. If so, an error will be returned.

## Motivation and Context

see https://github.com/orhun/rustypaste/discussions/241

I am opening this PR now, because I want to check, whether the test that fails in #246 also fails with this one.

Depending on which PR is merged first, a change will be required to add an `.await` to the `glob_match_file` call.

## How Has This Been Tested?

- manually
- cargo tests (new tests)
- fixtures (new tests)

## Changelog Entry

````
### Changed

- Return an error when a file with the same name already exists on the server
````

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -908,6 +908,63 @@ mod tests {
     }
 
     #[actix_web::test]
+    async fn test_upload_same_filename() -> Result<(), Error> {
+        let mut config = Config::default();
+        config.server.upload_path = env::current_dir()?;
+
+        let app = test::init_service(
+            App::new()
+                .app_data(Data::new(RwLock::new(config)))
+                .app_data(Data::new(Client::default()))
+                .configure(configure_routes),
+        )
+        .await;
+
+        let file_name = "test_file.txt";
+        let header_filename = "fn_from_header.txt";
+        let timestamp = util::get_system_time()?.as_secs().to_string();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/server.rs [hunk 1]
```diff
@@ -907,6 +907,63 @@ mod tests {
...
+            get_multipart_request(&timestamp, "file", file_name)
+                .insert_header((
+                    header::HeaderName::from_static("filename"),
+                    header::HeaderValue::from_static("fn_from_header.txt"),
+                ))
+                .to_request(),
+        )
+        .await;
+        assert_eq!(StatusCode::INTERNAL_SERVER_ERROR, response.status());
+        assert_body(response.into_body(), "file already exists\n").await?;
```
orhun, line 955: It would be nice to return [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) for cases like this but I guess it needs some refactoring since we return `IoResult` from `store_file`. I will create a separate issue for it.

orhun (approved):
LGTM!

PR #259, opened 2024-03-08
Test CI

tessus:
# Testing CI

Currently the same as #246

First theory: something is cached on gh and for some reason 246 can't get out of this state

2nd theory: default expiry has to be set explicitly in tests

3rd theory: tarpulin has a side effect


tessus:
Ok, I think we have to contact the tarpulin devs.

tessus:
Closing this again, after figuring out what the issue is.

/ref https://github.com/xd009642/tarpaulin/issues/1488

orhun:
Thanks for investigating this!

tessus:
You're welcome.

PR #262, opened 2024-03-08
refactor(server): use more specific HTTP status codes

### LINKED ISSUE #261: Return appropriate HTTP status code instead of 500
There are some cases where we return 500 instead of an appropriate HTTP status code. We should refactor them to make the API more idiomatic/conventional.

One example is from https://github.com/orhun/rustypaste/pull/258#discussion_r1517736731 

https://github.com/orhun/rustypaste/blob/4987cfe5e55e5fc6165fd8392b3c2396f6d34437/src/server.rs#L954-L955

We should use `409 Conflict` there instead of 500 internal server error.

tessus:
## Description

This change returns a more appropriate status code in 2 cases instead of a generic 500 code:

- UnsupportedMediaType (if the mime type is on the blacklist)
- Conflict (if the file already exists)

## Motivation and Context

closes #261 

## How Has This Been Tested?

- cargo test
- fixtures

## Changelog Entry

````
### Changed

- Return appropriate HTTP status code instead of 500
````

## Types of Changes

- [ ] Other 

## Checklist:

- [x] I have added (in this case changed) tests to cover my changes.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -951,15 +951,10 @@ mod tests {
                 .to_request(),
         )
         .await;
-        assert_eq!(StatusCode::INTERNAL_SERVER_ERROR, response.status());
+        assert_eq!(StatusCode::CONFLICT, response.status());
         assert_body(response.into_body(), "file already exists\n").await?;
 
         fs::remove_file(header_filename)?;
-        let serve_request = TestRequest::get()
-            .uri(&format!("/{header_filename}"))
-            .to_request();
-        let response = test::call_service(&app, serve_request).await;
-        assert_eq!(StatusCode::NOT_FOUND, response.status());
 
         Ok(())
     }
--- a/fixtures/test-file-upload-same-filename/test.sh
+++ b/fixtures/test-file-upload-same-filename/test.sh
@@ -13,6 +13,8 @@ run_test() {
   test "$content" = "$(curl -s $file_url)"
   file_url=$(curl -s -F "file=@file" -H "filename:fn_from_header.txt" localhost:8000)
   test "$file_url" = "file already exists"
+  status_code=$(curl -s -F "file=@file" -H "filename:fn_from_header.txt" -w "%{response_code}" -o /dev/null localhost:8000)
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/server.rs [hunk 1]
```diff
@@ -951,15 +951,10 @@ mod tests {
...
-        assert_eq!(StatusCode::INTERNAL_SERVER_ERROR, response.status());
+        assert_eq!(StatusCode::CONFLICT, response.status());
         assert_body(response.into_body(), "file already exists\n").await?;
 
         fs::remove_file(header_filename)?;
-        let serve_request = TestRequest::get()
-            .uri(&format!("/{header_filename}"))
-            .to_request();
-        let response = test::call_service(&app, serve_request).await;
-        assert_eq!(StatusCode::NOT_FOUND, response.status());
```
orhun, line 962: Why this is removed?

tessus: because we have that in another test. it's a duplicate and not necessary

tessus: see `test_upload_file_override_filename`

tessus: and `test_upload_file`

src/server.rs
```diff
@@ -951,15 +951,10 @@ mod tests {
                 .to_request(),
         )
         .await;
-        assert_eq!(StatusCode::INTERNAL_SERVER_ERROR, response.status());
+        assert_eq!(StatusCode::CONFLICT, response.status());
```
orhun, line 954: Do you think it makes sense to also update the related test fixtures to assert the status codes?

tessus: yep, I thought about it. but curl does not print out the status code without having to parse either the verbose output or the result of `-I`.

Although I remember there was some way. I have to check my notes. Maybe I can find it.

tessus: I'll look into it later tonight and add it to the fixtures.

tessus: done

orhun (approved):
Thanks!

Issue #312, opened 2024-06-27
paste with random names collide 


CvH:
We have several pastes that were generated with the same random name (running rustypaste 0.14.2 at ubuntu). 

config
```
[paste]
random_url = { type = "petname", words = 2, separator = "-" }
```

The command that creates that pastes for us is 
```
#!/bin/sh
cat "$@" | curl https://server.com -F 'file=@-' -H "Authorization: Bearer ${token}"
```

example
```
root@server:/rustypaste/upload# ls -lah lasting-giraffe.log*
-rw-r--r-- 1 rustypaste rustypaste 214K Feb  3 02:29 lasting-giraffe.log.1727967488761
-rw-r--r-- 1 rustypaste rustypaste 425K Mar 10 14:07 lasting-giraffe.log.1731119760881
root@server:/rustypaste/upload# ls -lah eager-sponge.log.*
-rw-r--r-- 1 rustypaste rustypaste 705K Jan 14 23:34 eager-sponge.log.1726315416603
-rw-r--r-- 1 rustypaste rustypaste 3.5K Jun 22 15:36 eager-sponge.log.1740110740452
```

Now while creating the paste it returns https://server.com/eager-sponge.log that points to the old log (that is not identical) and the new paste is not accessible at all.

I searched the issues but didn't found something similar sounding, is this a know problem and maybe solved with 0.15.0 ?

tessus:
All these issues should be fixed with 0.15.0. please see the release notes at https://github.com/orhun/rustypaste/releases/tag/v0.15.0

CvH:
tx, already updated
I have a look if this is fixed 👍

PR #325, opened 2024-08-02
chore(deps): bump petname to 2.0.2

### LINKED ISSUE #324: Bump petname dependency
See <https://github.com/allenap/rust-petname?tab=readme-ov-file#upgrading-from-1x>

tessus:
## Description

This change also uses the large word list due to #312. 
fixes #324

## Motivation and Context

see #324 and #312

## How Has This Been Tested?

- cargo test
- fixtures

## Changelog Entry

````
### Changed

- bump petname to 2.0.2 and use large word list
````

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/src/random.rs
+++ b/src/random.rs
@@ -33,7 +33,7 @@ impl RandomURLConfig {
                     self.words.unwrap_or(2),
                     self.separator.as_deref().unwrap_or("-"),
                 )
-                .expect("no names"),
+                ?,
             RandomURLType::Alphanumeric => rand::thread_rng()
                 .sample_iter(&Alphanumeric)
                 .take(self.length.unwrap_or(8))
--- a/src/random.rs
+++ b/src/random.rs
@@ -28,12 +28,10 @@ impl RandomURLConfig {
             return None;
         }
         Some(match self.type_ {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/random.rs [hunk 1]
```diff
@@ -27,10 +28,12 @@ impl RandomURLConfig {
...
-            RandomURLType::PetName => petname::petname(
-                self.words.unwrap_or(2),
-                self.separator.as_deref().unwrap_or("-"),
-            ),
+            RandomURLType::PetName => petname::Petnames::large()
+                .generate_one(
+                    self.words.unwrap_or(2),
+                    self.separator.as_deref().unwrap_or("-"),
+                )
+                .expect("no names"),
```
orhun, line 36: We can return from here and let the callee of this function handle the `None` case.

```suggestion
                ?,
```

tessus: I thought about it first, but I think it makes sense to panic here. This should never happen. The crate comes with a word list, so if None is returned, something is seriously wrong with the crate.
On the other side, a panic is prod is never useful. Letting the callee handle the error is also not a bad idea.

orhun (approved):
LGTM, thanks!

Issue #326, opened 2024-08-04
Host a public instance


orhun:
See #279 

I'm looking for a new cloud provider to host a public instance of `rustypaste`. Maybe spinning a Docker container with memory/storage limits in a private VPS would be sufficient.

orhun:
>  Do you want to remove the shuttle code, since shuttle is not very reliable?

Yes, I think it is useless at this point.

> Maybe you can get a cloud provider to sponsor a mini instance with limited traffic/storage.

Who should we reach out? Any ideas?

tessus:
> Yes, I think it is useless at this point.

I can have a look.

> Who should we reach out? Any ideas?

Hmm, AWS provided mini instances for free at one point. I am not sure, if they still do.
I believe Hetzner is quite popular in Europe. Maybe they have demo/test instances available for free. One could even share such an instance with others who also want to run a demo. Maybe they have something like this.

orhun:
I used AWS for a couple of years, I'd decide against it. Hetzner sounds good!

tessus:
Just read an article that Lidl is building a cloud system for and in Europe. Maybe give them a ping.

orhun:
Brought this up to Arch Linux: https://lists.archlinux.org/archives/list/arch-dev-public@lists.archlinux.org/thread/Y7KUOXXVVMXEAVFUPK4I5GXB6J55CIF6/

torrefatto:
You could try [koyeb](https://www.koyeb.com/), we have free instances (not stateful).

ogarcia:
@orhun [Oracle cloud](https://cloud.oracle.com) offers up to 4 machines [free of charge](https://docs.oracle.com/es-ww/iaas/Content/FreeTier/freetier_topic-Always_Free_Resources.htm#Details_of_the_Always_Free_Compute_instance__a1_flex) at all times. They don't have a huge amount of disk (47Gb), but for a basic pastes service in demo mode it should be enough. Besides, in Oracle to switch to “payment mode” you have to enable it deliberately, so it is impossible to get out of the free mode (unlike Amazon and similar that at the slightest you are already being charged).

Another option is to pay for a [Kimsufi server at OVH](https://eco.ovhcloud.com), they are not very expensive and that's what I do with _lesma_.

PR #328, opened 2024-08-05
refactor(server): remove shuttle code and deps

tessus:
## Description

This change removes the code for shuttle and also the gh workflow for shuttle deployment.

## Motivation and Context

see #326 

## How Has This Been Tested?

- cargo test
- fixtures

## Changelog Entry

````
### Removed

- shuttle code and deps
````

## Types of Changes

- [ ] Other 

## Checklist:

### CODE REVIEW

**Comment by orhun** [comment]
We can also remove the shuttleapp.rs references from `config.toml` and `examples/` :)

**Comment by tessus** [comment]
Good point. Will do.

**Review by orhun — APPROVED** [review]

File: config.toml [hunk 1]
```diff
@@ -3,7 +3,6 @@ refresh_rate = "1s"
 
 [server]
 address = "127.0.0.1:8000"
-#url = "https://rustypaste.shuttleapp.rs"
```
orhun, line 6: Second thought, since this config is not documented elsewhere maybe we can leave this in as a comment but change to URL to https://example.com or something like that.

What do you think?

tessus: yep, this probably makes sense. But I'd use something like:

```
https://paste.example.com
or
https://example.com/paste
```

In config.toml and examples/, right? Please let me know which one of the 2 above is better.

orhun: yup, sounds good. I think `config.toml` is better.

tessus: I meant one of the 2 from the URLs. ;-) I think we can add it to config.toml and to the examples/html_form.toml.

orhun: Ah, let's go with `paste.example.com` with both files then.

PR #335, opened 2024-08-18
feat: support `max_upload_dir_size` config

### LINKED ISSUE #282: Support a total size limit for the upload directory
It would be cool to have a configuration option for the total size of the upload folder. This would help with public instances in terms of avoiding a storage attack where a single user uploads a bunch of big/small files.

This new option might look like (or something similar):

```toml
[server]
max_uploads = "1G"
```

Narayanbhat166:
## Description

This Pull Request adds a feature to limit the uploads if the `./uploads` dir is full. The size of `./uploads` dir can be set by the config key `max_upload_dir_size` under the server config.

```toml
[server]
max_upload_dir_size = "100G"
```

## Motivation and Context

This is required to restrict the number / size of uploads that can be made to the server when hosted publicly.
This fixes #282 

## Changelog Entry

- Add a check for the max size of uploads directory

## Types of Changes

## Checklist:


src/config.rs
```diff
@@ -43,6 +43,8 @@ pub struct ServerConfig {
     pub max_content_length: Byte,
     /// Storage path.
     pub upload_path: PathBuf,
+    /// Maximum upload directory size.
+    pub max_uploads: Option<Byte>,
```
tessus, line 47: I would use a more distinct name. `max_uploads` sounds like how many uploads one can make.

Why don't use:

```suggestion
    pub max_upload_dir_size: Option<Byte>,
```

orhun: Yup, this change makes sense to me 👍🏼

Cargo.toml
```diff
@@ -46,6 +46,7 @@ tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uts2ts = "0.4.1"
 path-clean = "1.0.1"
+fs_extra = "1.3.0"
```
orhun, line 49: Do we require this dependency? We can simply add a function for traversing the directory and getting the size, which is essentially what `get_size` is doing: <https://docs.rs/fs_extra/latest/src/fs_extra/dir.rs.html#786-816>

fixtures/test-server-upload-dir-limit/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+setup() {
+  truncate -s 9KB bigfile1 bigfile2 bigfile3
+}
+
+run_test() {
+  result=$(curl -s -F "file=@bigfile1" localhost:8000)
+  result=$(curl -s -F "file=@bigfile2" localhost:8000)
```
orhun, line 9: Can you also check that these two files are uploaded? Adding a single check before attempting to upload `bigfile3` would be enough.

orhun (commented):
Thanks! Had some comments:

src/config.rs
tessus: Ah, I aee that you actually used my suggested name before. Cool. Thanks for switching it back.

orhun:
Let's get this over the finish line soon if possible :)

orhun:
Ping :)

tessus:
@Narayanbhat166 can you have a look and make the requested changes?

orhun (approved):

tessus:
@orhun I was just wondering what kind of perf impact this has. Isn't this similar to the `duplicates=false` issue? Every time there is an upload, all files have to be touched. Yes, getting the size of all files is probably a lot faster than calculating the hash for each file, but still. Patterns like this make me nervous, coming from a perf background.

orhun:
Yeah, probably the performance will be degrade in a noticeable way with this feature enabled.

I think we need to add benchmarks using [hyperfine](https://github.com/sharkdp/hyperfine) to get an idea about that. I did something similar for Zig in https://github.com/orhun/zig-http-benchmarks

PR #339, opened 2024-08-22
feat(server): add support for multiple auth tokens via env vars

### LINKED ISSUE #338: Loading Auth Tokens from a file
I am using [agenix](https://github.com/ryantm/agenix) to encrypt my secrets. Agenix only exposes the path to the decrypted file, and not the content itself, necessitating to give an application the path to the file at runtime.

Since rustypaste either loads the auth token from the `AUTH_TOKEN` env var or the configuration file, I find myself unable to pass multiple auth tokens to the server.

I'd be interested in adding the parsing of an authentication file at server start, this authentication file could be provided via an `AUTH_FILE` env var, which holds the path, and the file itself could have a newline separated list of authentication tokens.

If there is a different or better method of implementing this feature I'd gladly hear about it.

Nydragon:
## Description

Add support for reading a list of newline separated tokens from a file
which path is being supplied through the following environment
variables:

- AUTH_FILE
- DELETE_FILE

Also add support for setting environment variables inside of fixtures through the optional `custom_env` function.

Also describe this new functionality in the readme.

## Motivation and Context

Need a way to supply multiple auth token akin to `AUTH_TOKEN` due to agenix behaviour.

closes #338

## How Has This Been Tested?

Added UTs for file content parsing. File reading has not received any UT, but has been manually tested.

### Added

- Multiple authentication and deletion token parsing via a file to which `AUTH_FILE` and `DELETION_FILE` point. This file needs to be a newline separated list of strings. Whitespace is discarded.

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/fixtures/README.md
+++ b/fixtures/README.md
@@ -7,7 +7,7 @@ This directory contains the [test fixtures](https://en.wikipedia.org/wiki/Test_f
 1. Build the project in debug mode: `cargo build`
 2. Execute the runner script in this directory: `./test-fixtures.sh`
 
-On `macOS` you need to have [coreutils](https://www.gnu.org/software/coreutils/) installed to run the script. 
+On `macOS` you need to have [coreutils](https://www.gnu.org/software/coreutils/) installed to run the script.
 The simplest way is to install it via [Homebrew](https://brew.sh/): `brew install coreutils`
 
 ### Adding new fixtures
@@ -27,6 +27,11 @@ test-file-upload/
 
 ```sh
 #!/usr/bin/env bash
+
+# Optional
+custom_env() {
+  # setting environment variables such as AUTH_TOKEN or AUTH_FILE
+}
 
 setup() {
   # preparation
--- a/fixtures/README.md
+++ b/fixtures/README.md
@@ -30,7 +30,7 @@ test-file-upload/
 
 # Optional
 custom_env() {
-  # setting environment variables such as AUTH_TOKEN or AUTH_FILE
+  # setting environment variables such as AUTH_TOKEN or AUTH_TOKENS_FILE
 }
 
 setup() {
--- a/fixtures/test-fixtures.sh
+++ b/fixtures/test-fixtures.sh
@@ -9,11 +9,13 @@ NC='\033[0m'
 run_fixture() {
   cd "$FIXTURE_DIR/$1" || exit 1
   source "test.sh"
+  type custom_env &>/dev/null && custom_env
   NO_COLOR=1 CONFIG=config.toml "$PROJECT_DIR/target/debug/rustypaste" &
   SERVER_PID=$!
   trap 'kill -9 "$SERVER_PID" && wait "$SERVER_PID" 2> /dev/null' RETURN
   sleep 1
-  ( set -e;
+  (
+    set -e
     setup
     run_test
     teardown
@@ -24,7 +26,9 @@ run_fixture() {
 
 main() {
   find * -maxdepth 0 -type d -print0 | while IFS= read -r -d '' fixture; do
-    run_fixture "$fixture"
+    # Since we are creating a subshell, all environment variables created by custom_env will be lost
+    # Return code is preserved
+    (run_fixture "$fixture")
     exit_status=$?
     if [ "$exit_status" -eq 0 ]; then
       echo -e "[${GREEN}ok${NC}] $fixture"
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,10 +1,11 @@
 use crate::mime::MimeMatcher;
 use crate::random::RandomURLConfig;
-use crate::{AUTH_TOKEN_ENV, DELETE_TOKEN_ENV};
+use crate::{AUTH_TOKENS_FILE_ENV, AUTH_TOKEN_ENV, DELETE_TOKENS_FILE_ENV, DELETE_TOKEN_ENV};
 use byte_unit::Byte;
 use config::{self, ConfigError};
 use std::collections::HashSet;
 use std::env;
+use std::fs::read_to_string;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
@@ -162,6 +163,21 @@ impl Config {
                 if let Ok(env_token) = env::var(AUTH_TOKEN_ENV) {
                     tokens.insert(env_token);
                 }
+                if let Ok(env_path) = env::var(AUTH_TOKENS_FILE_ENV) {
+                    match read_to_string(&env_path) {
+                        Ok(s) => {
+                            s.lines().filter(|l| !l.trim().is_empty()).for_each(|l| {
+                                tokens.insert(l.to_string());
+                            });
+                        }
+                        Err(e) => {
+                            error!(
+                                "failed to read tokens from authentication file ({env_path}) ({e})"
+                            );
+                        }
+                    };
+                }
+
                 tokens
             }
             TokenType::Delete => {
@@ -170,6 +186,20 @@ impl Config {
                 if let Ok(env_token) = env::var(DELETE_TOKEN_ENV) {
                     tokens.insert(env_token);
                 }
+
+                if let Ok(env_path) = env::var(DELETE_TOKENS_FILE_ENV) {
+                    match read_to_string(&env_path) {
+                        Ok(s) => {
+                            s.lines().filter(|l| !l.trim().is_empty()).for_each(|l| {
+                                tokens.insert(l.to_string());
+                            });
+                        }
+                        Err(e) => {
+                            error!("failed to read deletion tokens from file ({env_path}) ({e})");
+                        }
+                    };
+                }
+
                 tokens
             }
         };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,5 +41,11 @@ pub const CONFIG_ENV: &str = "CONFIG";
 /// Environment variable for setting the authentication token.
 pub const AUTH_TOKEN_ENV: &str = "AUTH_TOKEN";
 
+/// Environment variable for the path to a file containing multiple authentication token.
+pub const AUTH_TOKENS_FILE_ENV: &str = "AUTH_TOKENS_FILE";
+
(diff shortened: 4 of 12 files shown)


tessus (approved):
Thanks for this PR. LGTM (unless I missed some Rust specific optimization I am not aware of, since I am not a Rust expert).

Nydragon:
> Oh, could you add a test to the fixtures as well?

Sure! I didn't see an option to set env vars tho... Did i miss something?

tessus:
> I was wondering if it would be better to `.trim()` the entire file or the individual lines after splitting (current behaviour).

One would have to run perf tests for the difference, but I doubt it is huge for a small to medium sized file with tokens.
Personally I like the current behavior better, since you took care of a few edge cases as well.

The only workflow optimization I can think of is that the file is not read and parsed every time a request is made, but only when the config is read or reloaded via `refresh_rate`.

If you want to try that, please go ahead.

Nydragon:
> > I was wondering if it would be better to `.trim()` the entire file or the individual lines after splitting (current behaviour).
> 
> One would have to run perf tests for the difference, but I doubt it is huge for a small to medium sized file with tokens. Personally I like the current behavior better, since you took care of a few edge cases as well.
> 
> The only workflow optimization I can think of is that the file is not read and parsed every time a request is made, but only when the config is read or reloaded via `refresh_rate`.
> 
> If you want to try that, please go ahead.

That'd make definitely sense! I was a bit surprised to see that `get_tokens` is called on every request. Why is that? Wouldn't it be enough to just call the function periodically according to `refresh_rate`?

tessus:
> That'd make definitely sense! I was a bit surprised to see that `get_tokens` is called on every request. Why is that? Wouldn't it be enough to just call the function periodically according to `refresh_rate`?

This is a good question. I think it developed this way. At the beginning there was only one token in the config file and one env var. I added the multiple tokens and delete tokens, and then it was refactored by somebody else to be used as a middleware with web grants.
In all these stages only the config was referenced (not the file read) and the ENV vars read, thus no real overhead.

src/config.rs
```diff
@@ -170,6 +180,16 @@ impl Config {
                 if let Ok(env_token) = env::var(DELETE_TOKEN_ENV) {
                     tokens.insert(env_token);
                 }
+
+                if let Ok(env_path) = env::var(DELETE_FILE_ENV) {
+                    match read_to_string(&env_path) {
+                        Ok(s) => push_token_file(s, &mut tokens),
```
orhun, line 186: No need for a function call:

```suggestion
                        Ok(s) => {
                            s.lines().filter(|l| !l.trim().is_empty()).for_each(|l| {
                                tokens.insert(l.to_string());
                            });
                        }
```

src/config.rs
```diff
@@ -170,6 +180,16 @@ impl Config {
...
+                        Err(e) => {
+                            error!("Failed reading from deletion file {env_path}.\n{e}");
```
orhun, line 188:
```suggestion
                            error!("failed to read deletion tokens from file ({env_path}) ({e})");
```

src/config.rs
```diff
@@ -162,6 +163,15 @@ impl Config {
                 if let Ok(env_token) = env::var(AUTH_TOKEN_ENV) {
                     tokens.insert(env_token);
                 }
+                if let Ok(env_path) = env::var(AUTH_FILE_ENV) {
+                    match read_to_string(&env_path) {
+                        Ok(s) => push_token_file(s, &mut tokens),
```
orhun, line 168: No need for a function call.

```suggestion
                        Ok(s) => {
                            s.lines().filter(|l| !l.trim().is_empty()).for_each(|l| {
                                tokens.insert(l.to_string());
                            });
                        }
```

src/config.rs
```diff
@@ -203,6 +223,18 @@ impl Config {
...
+        .split('\n')
+        .filter_map(|s| match s.trim() {
+            "" => None,
+            s => Some(s),
+        })
+        .for_each(|t| {
+            tokens.insert(t.to_string());
+        });
+}
+
```
orhun, line 237: Not needed.

orhun (commented):
Looks good! We can simplify it a bit :)

orhun:
> are you ok with adding another hook to add env vars to the fixtures?

Hmm, yeah, if it is not much of a hassle. Bash is annoying.

>  I was a bit surprised to see that get_tokens is called on every request. Why is that? 

We use a library called `actix_web_grants` which works by adding a middleware and then you can simply protect your endpoints with an annotation.

https://github.com/orhun/rustypaste/blob/1d5a9c60044ab0ece53868eccbe7cf011d0c5633/src/server.rs#L417

https://github.com/orhun/rustypaste/blob/1d5a9c60044ab0ece53868eccbe7cf011d0c5633/src/server.rs#L153

Implemented in #199 

That is why `extract_tokens` is called on every request since there is a guard. We can maybe cache some parts to make it more performant.

> Wouldn't it be enough to just call the function periodically according to refresh_rate?

That means we need to store the tokens somewhere (probably in memory) and that's not how the grants library works.

src/config.rs
Nydragon: I originally created a function to be able to write a UT for the parsing, is just a fixture fine then ?

Nydragon:
Alright, thank you for the explanation, should I then leave out the improvements mentioned by tessus (https://github.com/orhun/rustypaste/pull/339#issuecomment-2304652210)?

I'll also add the env handling to the fixture script.

src/config.rs
orhun: Yup, it should be enough.

Nydragon:
Should be all good now:

- added fixture for the file based authentication process
- added fixture for the file based deletion process
- added env var setting through the `rp_env` function, teardown is handled by putting `run_fixture` into a subshell
- applied code review requests

fixtures/README.md
```diff
@@ -28,6 +28,11 @@ test-file-upload/
 ```sh
 #!/usr/bin/env bash
 
+# Optional
+rp_env() {
```
orhun, line 32: I think it would be better if we rename this to `custom_env`.

fixtures/README.md
```diff
@@ -28,6 +28,11 @@ test-file-upload/
...
+  # setting env variables such as AUTH_TOKEN or AUTH_FILE
```
orhun, line 33: nit:
```suggestion
  # setting environment variables such as AUTH_TOKEN or AUTH_FILE
```

fixtures/test-fixtures.sh
```diff
@@ -9,11 +9,14 @@ NC='\033[0m'
 run_fixture() {
   cd "$FIXTURE_DIR/$1" || exit 1
   source "test.sh"
+  # Check if rp_env exists
```
orhun, line 12: it's clear, no need for a comment here :)

fixtures/test-fixtures.sh
```diff
@@ -24,7 +27,9 @@ run_fixture() {
 
 main() {
   find * -maxdepth 0 -type d -print0 | while IFS= read -r -d '' fixture; do
-    run_fixture "$fixture"
+    # Since we are creating a subshell, all env vars created by rp_env will be lost
+    # Return code is preserved
+    (run_fixture "$fixture")
```
orhun, line 32: Is there any side effects of this? Are we sure that the return code is preserved?

I'd test it manually with a failing fixture on purpose to see if the behavior is the same.

also nit:

```suggestion
    # Since we are creating a subshell, all environment variables created by rp_env will be lost
    # Return code is preserved
    (run_fixture "$fixture")
```

tessus: Interesting. I would have used `custom_env_teardown()` and added it to `run_fixture()` before retuning.

But if the subshell works, it's certainly a better option.

Nydragon: I echo'd the return value during my modifications and it did preserve the return code properly. As for any side effects I couldn't notice any.

<details><summary>Fixtures run normally:</summary>
<p>

image
</p>
</details> 

<details><summary>Modified `test` to fail</summary>
<p>

image

</p>
</details>

Nydragon:
Applied all requests

tessus (commented):
Great, thanks.

One last thing. Could you please update the README in the fixtures dir?

Can you please add the new function to the file format?

Nydragon:
I think I already did that: https://github.com/orhun/rustypaste/pull/339/files#diff-3374186a79a0a6316047cb823494f238bdc644489b1b709bf65c192b5c05183a

tessus:
Thanks, I missed it. All good then! I think it is ready to be merged.

src/lib.rs
```diff
@@ -41,5 +41,11 @@ pub const CONFIG_ENV: &str = "CONFIG";
...
 pub const AUTH_TOKEN_ENV: &str = "AUTH_TOKEN";
 
+/// Environment variable for the path to a file containing multiple authentication token.
+pub const AUTH_FILE_ENV: &str = "AUTH_FILE";
+
 /// Environment variable for setting the deletion token.
 pub const DELETE_TOKEN_ENV: &str = "DELETE_TOKEN";
+
+/// Environment variable for the path to a file containing multiple deletion token.
+pub const DELETE_FILE_ENV: &str = "DELETE_FILE";
```
orhun, line 51: I was about to merge this but then had an epiphany, `DELETE_FILE` sounds more like "delete this file" rather than "a file that contains the deletion tokens". I think we should grammatically fix this while we have the chance :) So I propose:

- `AUTH_FILE` -> `AUTH_TOKENS_FILE`
- `DELETE_FILE` -> `DELETE_TOKENS_FILE`

Thoughts? @Nydragon @tessus

Nydragon: Completely agree, actually chose that name at the beginning, but replaced it with the current one because it was shorter.

Will update that as soon as I'm back home!

tessus: Shorter names only make sense, if it has to be used all the time....

Nydragon: Changed the names

orhun (approved):
Perfect, thanks!

Issue #336, opened 2024-08-18
upload limit is ignored


tessus:
I have set `max_content_length = "50MB"`, but whenever I upload a file larger than 50MB, the entire file is uploaded and only then I get the error message that the limit is exceeded.

# rpaste

https://github.com/user-attachments/assets/a4c88e80-980b-4614-89dd-a679aca1ab91

# curl

https://github.com/user-attachments/assets/b427758e-57b1-4d94-ae15-3e14bc75de44

orhun:
The middleware was implemented in #53 and yes, it turns out it is broken. I just reproduced this issue.

What would be the way of going forward with fixing this? Should we check HEAD requests, or do something in the middleware? Any ideas?

tessus:
> What would be the way of going forward with fixing this?

At one point I read a comment (but I am not sure in which repo or issue/discussion) that there is now a native way with the new versions of `actix` and/or `actix-multipart` which should render the middleware obsolete.

I've been looking for that comment frantically, but unfortunately haven't found it yet.

orhun:
It would be the best way of moving forward if something like that exists. Custom implementations are not always infallible, like we saw in this case :D

tessus:
I think I found it: https://stackoverflow.com/a/78399675

orhun:
That's superb, can you try it out?

One thing to note that we probably won't be able to change this on the fly since it's configured while the App is constructed.

tessus:
Not sure when I get to it, but yes, I would like to.

tessus:
I tried it, but unfortunately no difference. Maybe there's an issue with the `Byte` arithmetic. I am quite puzzled by this.
When I get to it, I'll open a draft PR and you can have a look at my attempt to make it woek with the "new" way of doing it according to the stackoverflow comment. Maybe I fucked it up though...

orhun:
Yup, shoot me a draft whenever you can!

tessus:
This (#360) was supposed to work, but it doesn't. I am not sure how to solve this. I can't believe nobody ever got this to work.

orhun:
We both seem quite puzzled by this. Maybe a gentle ping to the actix master @robjtede would help us out a bit 😊

PR #360, opened 2024-10-25
fix(server): upload limit

tessus:
## Description

Update code to limit the file upload.

## Motivation and Context

see #336
and https://stackoverflow.com/a/78399675

## How Has This Been Tested?

It does not seem to work.

## Types of Changes

- [ ] Other 

## Checklist:

Will be done after moving it out of draft status.


tessus:
I didn't adjust the tests, which is why the pipeline complains.

But I was testing manually and it doesn't work.

orhun:
Hmm, what's the status with this?

tessus:
This PR implements the limit as was shown at stackoverflow. You can build the app and test it. But it doesn't work. Same behavior as before, except that now you don't even get an error message after the file has been uploaded. Apparently limiting the fie upload is just not possible with this framework or every single information out there about how one can accomplish this is wrong.

You pinged the author here https://github.com/orhun/rustypaste/issues/336#issuecomment-2443504870 but nothing so far.

orhun:
I see. I might come back to this later then.

tessus:
Yea, this one is a draft and was just to show that it doesn't work as described in the stackoverflow article.
We can leave it for reference for now. Who knows maybe Rob will check it out... ;-)

PR #382, opened 2025-01-02
change: use fully locked system accounts

inglor:
systemd v257 supports and recommends fully locked system accounts[1]

[1]: https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html#u

## Description

use fully locked system accounts

## Motivation and Context

Recommendation from systmd v257 and above
https://github.com/systemd/systemd/blob/25a306f6c4d9b8596e25d7771802f005ebae91ba/NEWS#L822-L832

## How Has This Been Tested?

Not tested as this is downstream

### Changed
- use fully locked sytem acounts on systemd > v257

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/extra/systemd257+/rustypaste.sysusers
+++ b/extra/systemd257+/rustypaste.sysusers
@@ -1,0 +1,2 @@
+# use fully locked system account on systemd 257+ (note the !)
+u! rustypaste - "Minimal file upload/pastebin service" /var/lib/rustypaste


tessus (commented):
In theory this is a good change. However, what happens, if people use systemd <257?
Have you tested this with systemd <257?

tessus:
I can't recall whether comments are allowed in this file. If so, just add a comment below the current one to add the ! suffix if systemd >=257 is used.

Otherwise a README or a second file in the systemd dir could help. Or you create a second dir, e.g. `systemd257+` or something...

But the change as it is right now is not feasible. Maybe in 5 years or so, but doing this a few weeks after 257 has been released is way too early.

inglor:
> Otherwise a README or a second file in the systemd dir could help. Or you create a second dir, e.g. `systemd257+` or something...
> 
made it a new directory

> But the change as it is right now is not feasible. Maybe in 5 years or so, but doing this a few weeks after 257 has been released is way too early.
>
I don't think it's gonna be in 5 years but I don't mind..

orhun:
I think it's fine now since the directory is named `systemd257+` which clearly communicates that this belongs to `>257`.

> just add a comment below the current one

This is a good point though, maybe we can add a comment about this line, e.g. `# use fully locked system account` or such, if that file supports comments.

Otherwise this LGTM.

orhun (approved):

tessus:
I checked the docs: _Empty lines and lines beginning with the "#" character are ignored, and may be used for commenting._

Please add the comment (or I can add it) an then we can merge (orhun has approved the PR).

orhun (approved):

Issue #391, opened 2025-01-28
remote image or access to remote image not stable


tessus:
2 of the tests use a remote image `https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg`.

But it happens now and then that the tests fail:

```
---- paste::tests::test_paste_data stdout ----
thread 'paste::tests::test_paste_data' panicked at src/paste.rs:537:9:
assertion `left == right` failed
  left: "70ff72a2f7651b5fae3aa9834e03d2a2233c52036610562f7fa04e089e8198ed"
 right: "06aaaa43a11f81b8a91e584fda70f19e6d69b53f67e24019a37937941f146805"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- server::tests::test_upload_remote_file stdout ----
thread 'server::tests::test_upload_remote_file' panicked at src/server.rs:1134:9:
assertion `left == right` failed
  left: "70ff72a2f7651b5fae3aa9834e03d2a2233c52036610562f7fa04e089e8198ed"
 right: "229f7bb9dde15a3dac2a0a06cdfd5f49f0e3b6267d1a64c5f24923f03b6ab0f7"
```

It is hard to say what exactly the issue is. Maybe we run into a rate limit.
I am thinking about replacing the remote URL and the image.

- [x] add small image to this repo #392
- [x] use image (`https://raw.githubusercontent.com/...`) in tests #394

PR #392, opened 2025-01-28
chore(test): add image for tests

tessus:
## Description

see #391

This image has only 911 bytes. (First I thought about using the existing rp logo, but it's 219k - or 165k after optimization)

## Motivation and Context

2 tests sometimes fail. This change is a pre-req for changing the tests in the next PR.
This file has to exist in the repo, so that it can be accessed via `https://raw.githubusercontent.com/...`


tessus:
@orhun can you please approve so that I can create the next PR? I can't create a PR with changes to tests that access an image that doesn't exist yet. ;-)

img/README.md
```diff
@@ -0,0 +1,3 @@
+The image `rp.png` is used for tests and must not be changed.
+
+SHA256 (rp.png) = 3b5eeeee7a7326cd6141f54820e6356a0e9d1dd4021407cb1d5e9de9f034ed2f
```
orhun, line 3: I would rather we remove this file and rename `rp.png` to `rp_test_{sha256}.png` along with the hash maybe?

tessus: Sure, give me a sec. The reason for the README was to clarify that removing the image file might not be a good idea. haha. But since there are not a thousand people working in this project, we should be good.

A file rename coming up...

tessus: done

orhun (approved):

PR #409, opened 2025-02-28
refactor: prepare for rust edition 2024

tessus:
## Description

Use `unsafe` blocks where required by edition 2024.

## Motivation and Context

Edition 2024 requires setting and removing env vars to be in an `unsafe` block.
When switching to edition 2024 in the future, a `cargo fmt` will be required as well.

## How Has This Been Tested?

cargo test
fixtures

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -243,7 +243,9 @@ mod tests {
     #[test]
     fn test_parse_config() -> Result<(), ConfigError> {
         let config_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("config.toml");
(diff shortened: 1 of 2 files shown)


orhun:
Ugh... this is not great. Do you know any way of setting these environment variables in a safe way?

tessus:
I am afraid there is not. see https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html

Maybe there's a crate for env handling, but even then one would have to check how the env vars are actually processed.

orhun (approved):

PR #414, opened 2025-03-07
ci(lychee): add `.lycheeignore` with new links

tessus:
For a while now, 2 links run into a timeout sporadically. I've exluded `repology.org` from the check.

```
[TIMEOUT] https://repology.org/badge/vertical-allrepos/rustypaste.svg | Timeout
[TIMEOUT] https://repology.org/project/rustypaste/versions | Timeout
```

Additionally according to the lychee documentation the exclude argument should be enclosed in single quotes.
I have also escaped the `.`, since we are using a regex.

We could also use a `.lycheeignore` file. e.g. it would make it easier to add complete links to that file, than changing the ci workflow file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,6 @@ jobs:
       - name: Check the links
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --exclude "example.com|site.com|localhost|awesome.txt|x.txt" -v *.md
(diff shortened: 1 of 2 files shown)


orhun:
> We could also use a .lycheeignore file. e.g. it would make it easier to add complete links to that file, than changing the ci workflow file.

Let's do that instead! (in this PR or separate)

tessus:
Done

orhun (approved):

PR #438, opened 2025-04-24
fix(fixtures): script and one test

tessus:
If the `test-fixtures.sh` script uses an argument, the test will always pass no matter what.
This change fixes the issue.

Additionally, the `test-server-upload-dir-limit` failed for multiple reasons:

- truncate on macOS does not understand `KB`
- curl returns an exit code 23 if you try to output a binary file to the terminal
- bash prints a warning if it encounters a null byte (this does not fail the test, but is still annoying)

P.S.: I still do not understand how the upload_dir_limit test could have ever passed in the pipeline.

### DIFF
--- a/fixtures/test-fixtures.sh
+++ b/fixtures/test-fixtures.sh
@@ -41,14 +41,16 @@ process_fixture() {
 
 main() {
   # If arguments are passed, run only those fixtures
-  [ $# -ne 0 ] && for fixture in "$@"; do
-    process_fixture "$fixture"
-  done && exit 0
-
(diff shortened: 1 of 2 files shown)


orhun (approved):
LGTM, can you split these changes into two commits so that we can rebase and merge?

I personally don't like having two separate fixes in a single commit.

tessus:
Yep, I thought about it at first, but then I thought the PRs are squashed anyway. But you are correct, they don't have to be.
I'll create 2 commits.

tessus:
Done.

PR #541, opened 2026-04-12
chore(deps): update petname and rand

### LINKED ISSUE #540: RUSTSEC-2026-0097: Rand is unsound with a custom logger using `rand::rng()`
> Rand is unsound with a custom logger using `rand::rng()`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unsound                |
| Package             | `rand`                      |
| Version             | `0.9.2`                   |
| URL                 | [https://github.com/rust-random/rand/pull/1763](https://github.com/rust-random/rand/pull/1763) |
| Date                | 2026-04-09                         |

It has been reported (by @lopopolo) that the `rand` library is [unsound](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#soundness-of-code--of-a-library) (i.e. that safe code using the public API can cause Undefined Behaviour) when all the following conditions are met:

- The `log` and `thread_rng` features are enabled
- A [custom logger](https://docs.rs/log/latest/log/#implementing-a-logger) is defined
- The custom logger accesses `rand::rng()` (previously `rand::thread_rng()`) and calls any `TryRng` (previously `RngCore`) methods on `ThreadRng`
- The `ThreadRng` (attempts to) reseed while called from the custom logger (this happens every 64 kB of generated data)
- Trace-level logging is enabled or warn-level logging is enabled and the random source (the `getrandom` crate) is unable to provide a new seed

`TryRng` (previously `RngCore`) methods for `ThreadRng` use `unsafe` code to cast `*mut BlockRng&lt;ReseedingCore&gt;` to `&amp;mut BlockRng&lt;ReseedingCore&gt;`. When all the above conditions are met this results in an aliased mutable reference, violating the Stacked Borrows rules. Miri is able to detect this violation in sample code. Since construction of [aliased mutable references is Undefined Behaviour](https://doc.rust-lang.org/stable/nomicon/references.html), the behaviour of optimized builds is hard to predict.

Affected versions of `rand` are `&gt;= 0.7, &lt; 0.9.3` and `0.10.0`.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2026-0097.html) for additional details.

tessus:
I've updated petname and rand. Both required code changes, which I have done.
Then I also updated other occurrences of rand via `cargo update`.

Unfortunately there are 2 packages that have rand 0.8.5 locked in. However, I don't think that we meet the conditions for an exploit.

closes #540

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,7 +73,7 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rand 0.9.2",
+ "rand 0.9.3",
  "sha1",
  "smallvec",
  "tokio",
@@ -89,7 +89,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.82",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -127,7 +127,7 @@ dependencies = [
  "parse-size",
  "proc-macro2",
  "quote",
- "syn 2.0.82",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


orhun (approved):
thanks!

PR #561, opened 2026-05-13
chore(release): prepare for v0.17.0

orhun:
closes #559


CHANGELOG.md
```diff
@@ -5,6 +5,85 @@ All notable changes to this project will be documented in this file.
...
+  "application/x-yaml",
+]
+
+hardening = true
+```
+
+1. `text_mime_overrides`: Additional MIME types to render as `text/plain` when serving files. This is useful for preventing XSS attacks when serving files that would otherwise be rendered as HTML or JavaScript in the browser.
+2. `hardening`: When set to `true`, the server will include security headers such as `X-Content-Type-Options: nosniff` and a basic `Content-Security-Policy` to mitigate XSS and other attacks.
+
+Remote URL uploads are now also validated to reject localhost and non-public IP addresses.
```
tessus: True, except the ip4 private ranges 10. 192.168. are allowed. Otherwise you could not eun it in an intranet

PR #563, opened 2026-05-29
feat(paste): improve compound extension detection using mime_guess

coko7:
## Description

Extracts the inline file extension parsing from `Paste::store_file` into a new
`util::get_extension_from_filename` helper.

The helper uses `mime_guess` to detect compound extensions (e.g. `tar.gz`): it
iterates candidate suffixes from longest to shortest and, when a known
compression/archive suffix is found (e.g. `gz`), also checks whether the immediately
preceding part is itself a recognized type (e.g. `tar`) — returning the combined form
(`tar.gz`) when it is.

Falls back to `Path::extension` for single-component or unknown
extensions.

## Motivation and Context

The previous parsing logic treated everything after the first dot as the extension, so a
file like `foo.bar.baz.txt` would produce `<random>.bar.baz.txt` in suffix mode despite
being a plain text file. The new helper uses `mime_guess` to identify only meaningful
extensions: single ones like `.txt`, and genuine compound formats like `.tar.gz` where
both components are recognized types.

## How Has This Been Tested?

- Existing `paste::tests::test_paste_data` covers `store_file` end-to-end for plain
  extensions, compound extensions (`foo.tar.gz`, `.foo.tar.gz`), dotfiles, and
  random-URL suffix mode.
- New `util::tests::test_get_extension_from_filename` unit-test to test the helper

## Changelog Entry

```
### Changed

- Extract file extension parsing into `util::get_extension_from_filename`.
  - Uses `mime_guess` to correctly detect compound extensions (e.g. `tar.gz`) while
    avoiding treating every dot-separated segment as part of the extension.
```

## Types of Changes

- [ ] Other 

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1435,9 +1435,9 @@ checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
 
 [[package]]
 name = "mime_guess"
-version = "2.0.4"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4192263c238a5f0d0c6bfd21f336a313a4ce1c450542449ca191bb657b4642ef"
+checksum = "f7c44f8e672c00fe5308fa235f821cb4198414e1c77935c1ab6948d3fd78550e"
 dependencies = [
  "mime",
  "unicase",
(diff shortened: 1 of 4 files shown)


orhun (approved):
I think this is fair, thanks for the PR!

tessus (approved):
Cool. This improves my previous extension handling kung fu and makes it less complicated. Nicely done.

Issue #578, opened 2026-06-24
Add list endpoint for shortened URL


Bit38:
Hi!
I started using this project a few months back. Rustypaste has an ability to list all uploaded files with the `/list` endpoint, but it does not support listing shortened URLs.  I believe such functionality should exist. I tried to figure out what would be the best way to implement this feature:
1. Custom endpoint something like: `/urls` - this is not as descriptive as `/list`
2. Nested endpoint: `/list/url` - no other endpoint is nested
3. Some specific header (or query param) for `/list` endpoint that will specify what *type* of list will be returned

I don't have much experience with rust, but in the following days I will try to create a draft PR(s)/implementation(s). Thank you!

tessus:
Hmm, I recall that this conversation came up a few months ago. see https://github.com/orhun/rustypaste-cli/issues/251

@orhun what's your take on https://github.com/orhun/rustypaste-cli/issues/251#issuecomment-4362087057

orhun:
Since this came up again, I think it makes sense to implement it.
I'd be fine with expanding the model and adding a file_type like it's proposed in the discussion. Or we could do nested URLs like it's originally proposed in the issue.

tessus:
@Bit38 any update on this? I am more than happy to review a PR.

Bit38:
I have implemented the core logic. I just haven't had a time to look at the tests yet. I think I will be able to work on them today and submit a PR today or tomorrow.

tessus:
No worries, no need to rush. Go at your pace. I just wasn't sure whether you are still interested, since there was no reply after Orhun's comment. All good.

PR #505, opened 2025-12-16
feat: add password-protected file uploads

anoldguy:
## Description

This PR adds password-protected file uploads to rustypaste. Users upload files with the `protected` form field, the server generates a random password and returns it alongside the URL, and downloads require that password via Authorization header.

I tried to follow the existing paste type patterns closely. Passwords are hashed with Argon2id and stored in sidecar files (`filename.txt.password`) alongside the uploaded content. This avoids extended attributes or database dependencies while keeping cleanup straightforward.

File operations use careful ordering to prevent unprotected content exposure. Creation stores password hashes before writing content files, and reverse that order for deletion. If either operation is interrupted, worst case is an orphaned password file with no exposed content. The cleanup routine removes these orphaned files when detected.

## Motivation and Context

Sharing sensitive files requires more protection than obscurity through random URLs. Password protection adds a basic authentication layer while keeping rustypaste's simplicity intact.

This implementation aims to fit naturally into the existing codebase rather than introduce new architectural patterns. The sidecar file approach mirrors how expiration timestamps are handled and makes cleanup deterministic.

## Design Decisions

**Storage location:** Protected files are stored in a dedicated `protected/` subdirectory, following the same pattern as oneshot and URL paste types. This provides natural isolation and simplifies the codebase by eliminating the need for special filtering logic in the list endpoint (my first approach).

### Manual Testing

Tested locally against the development server:

- Upload with `curl -F "protected=@file.txt"` returns URL and password
- Download with Bearer token succeeds: `curl -H "Authorization: Bearer <password>"`
- Download with Basic Auth succeeds: `curl -u "user:<password>"`
- All failure modes return 404 to prevent enumeration (missing auth, wrong password, or non-existent file all return identical responses)
- Cleanup removes both main file and password sidecar
- Orphaned password files are identified and removed

### Automated Tests

**Unit tests** (`src/password.rs`):
- `test_password_hashing` - Verifies hash generation and password verification against Argon2
- `test_password_file_path` - Tests path transformation logic for password sidecar files
- `test_password_file_path_invalid_utf8` - Validates error handling for paths with invalid UTF-8 sequences
- `test_store_and_verify_password` - Full roundtrip test covering storage, verification, and cleanup

**Integration tests** (`src/server.rs`):
- `test_upload_protected_file` - Tests upload/download flow with all authentication scenarios (missing auth, wrong password via Bearer/Basic, correct password via Bearer/Basic)
[...]

PR #587, opened 2026-07-09
feat: add item type to /list endpoint

Bit38:
## Description

The `/list` endpoint returned only *normal* files (`PasteType::File` and `PasteType::RemoteFile`). Now it returns also every other supported type and reports it using `item_type` field. Example response:
```json
 [
  {
    "file_name": "crunchiest-cristiano.toml",
    "file_size": 1956,
    "item_type": "file",
    "creation_date_utc": "2026-07-09 16:16:56",
    "expires_at_utc": "2026-07-09 16:26:56"
  }
]
``` 

## Motivation and Context

Solves https://github.com/orhun/rustypaste/issues/578

## How Has This Been Tested?

- Added test `server::tests::test_list_item_type`
- Manual testing on running instance

### Added
- Included item/file type in `/list` endpoint
  - `/list` endpoint now returns json with field `item_type` (possible values: `file`, `oneshot`, `url`, `oneshot_url`) indicating type of uploaded content.
  - All types of uploads are now included in the response.

## Types of Changes

- [ ] Other 

## Checklist:


src/server.rs
```diff
@@ -405,59 +407,80 @@ async fn list(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
...
+                        {
+                            return None;
+                        }
+                        Some(uts2ts::uts2ts(expiration / 1000).as_string())
+                    } else {
+                        None
+                    };
+                    Some(ListItem {
+                        file_name,
+                        // NOTE: For urls this will return url length. Not sure if desired behaviour
```
tessus: It is intended behavior. Otherwise what would you use for an url? Even if you were to grab the size of the retrieved webpage, what exactly would you add to it? html, css, js, ... Also, that value could change during the lifetime of that link.

Thus the only number that makes sense is either `0`, `null`, or the length of the link.

src/server.rs
```diff
@@ -405,59 +407,80 @@ async fn list(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
...
+                        creation_date_utc,
+                        expires_at_utc,
+                    })
                 })
             })
-        })
+            .collect())
+    };
+
+    let entries: Vec<ListItem> = PASTE_VARIANTS_LIST
```
tessus, line 542: Am I correct that all files are first added and then filtered based on `PASTE_VARIANTS_LIST`?
Because there's a PR open to add another type `ProtectedFile` that is purposely not returned by the list endpoint.

src/server.rs
```diff
@@ -789,6 +812,88 @@ mod tests {
...
+        let result: Vec<ListItem> = test::call_and_read_body_json(&app, request).await;
+
+        assert_eq!(result.len(), 4);
+
+        // Items returned from `/list` endpoint should be returned in this order:
+        // 1. PasteType::File
+        // 2. PasteType::Oneshot
+        // 3. PasteType::Url
+        // 4. PasteType::OneshotUrl
+        // NOTE: The test won't pass if order in `PASTE_VARIANTS_LIST` changes
```
tessus, line 948: I have to get back and check the code (unless Orhun beats me to it), but I can't recall whether we sort at all. e.g. by expiry.
By default I would not expect the `/list` endpoint to do any sorting at all (that can be done with jq afterwards), but if you need sorting I don't mind either. In that case we should agree on a default sorting. I am not saying the current one is not ok. I just wanted to put it out there.

Please ignore the following:

IDEA tessus /feat: add sorting to list endpoint

For future reference for myself: sorting the list endpoint might be a cool feature. e.g. `/list/expiry/desc`

tessus (commented):
Thanks a bunch for the PR. I will test it in the coming days, especially in combination with #505.

I've added a few inline comments. Maybe Orhun has some comments as well.

src/server.rs
Bit38: I iterate over `PASTE_VARIANTS_LIST`, read each variant's directory, collect its top-level files into a `Vec<ListItem>`, and flatten the result. The rest of the original logic is the same - I only added the `item_type` field and moved it into a closure.

From my understanding from glancing at the other PR, the `ProtectedFile` type should not be included in the response by default, as they are stored in their own subfolder.

tessus: > From my understanding from glancing at the other PR, the `ProtectedFile` type should not be included in the response by default, as they are stored in their own subfolder.

Yep, I had the same impression. I should be able to test it soon. I am travelling and my time's a bit limited, but I'll try my best.

orhun:
Looks good generally, let me know when this is ready for an additional review @tessus

tessus:
I have to wait until #589 is merged before I can fix the errors. 

Edit: fixed the PR number.

tessus:
I have been testing it and it looks fine so far. Protected files are not listed.
But I ran into another issue that has nothing to do with this PR: while the protected file tests pass in the code and the fixtures, I cannot retrieve any protected files. I have to run a few tests to figure out what the problem is.

I am a bit exhausted today so I will run the rest of my tests for this PR tomorrow (e.g. testing the cli app, ...).

In the meantime, maybe @orhun can decide what we should use in the `file_size` field for urls: `0`, `null`, or the length of the link (as it is currently).

orhun:
I think file_size should probably be null for URLs - that's the option that makes sense to me most.

src/server.rs
```diff
@@ -469,59 +471,79 @@ async fn list(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
...
+                        {
+                            return None;
+                        }
+                        Some(uts2ts::uts2ts(expiration / 1000).as_string())
+                    } else {
+                        None
+                    };
+                    Some(ListItem {
+                        file_name,
+                        file_size: metadata.len(),
```
tessus: Ok, let's use a JSON `null` for url and oneshoturl.

tessus (changes requested):
Please see my inline comment. I've tested everything and it works great.

When this is merged the cli needs an update as well for the pretty print of the list. But I'll take care of it as soon as this is merged. @orhun we have to make a few design decisions regarding the pretty print of the file list. I'll create a discussion later. 

Update: https://github.com/orhun/rustypaste-cli/discussions/257

tessus:
@orhun using file_size null for urls will break `rpaste -lp`, but I don't think that this is a problem. The pretty print is a convenience function in listing server files and thus requiring a new rpaste version is valid.

tessus:
Something else came to mind.

While I do understand that listing all types is nice, it also highlights a fact that might become more obvious. Deleting anything but files does not work, because it was never implemented.

The question now is whether to add this to this PR or create a new one. 

Update: I am more than happy to add this to this PR. I didn't mean @Bit38 has to implement this.

tessus:
I'll create a separate PR to allow the deletion of all item types. I think it makes more sense to separate these 2 features. I rather dislike to put too many features/fixes in the same PR (or commit).

@Bit38 let me know if you need help with the file_size `null` value for url and oneshot_url. Tip: you only have to change 2 lines (not consecutive) in the code.

src/server.rs
Bit38: done

src/server.rs
```diff
@@ -469,59 +471,82 @@ async fn list(config: web::Data<RwLock<Config>>) -> Result<HttpResponse, Error>
...
+                        }
+                        Some(uts2ts::uts2ts(expiration / 1000).as_string())
+                    } else {
+                        None
+                    };
+                    Some(ListItem {
+                        file_name,
+                        file_size: match item_type {
+                            PasteType::Url | PasteType::OneshotUrl => None,
+                            _ => Some(metadata.len()),
```
tessus:
```suggestion
                            PasteType::File | PasteType::Oneshot => Some(metadata.len()),
                            _ => None,
```

I am not sure, if it makes any difference. I have learned to put the happy path first (it is probably more likely that files are uploaded than urls) to help the compiler and the CPU with branch prediction, but it is also very possible that none of this matters anymore, because the compilers are so much more advanced these days.

Feel free to ignore my suggestion.

tessus (approved):
Awesome, thank you. LGTM.

@orhun I believe it is time for another review.

orhun (approved):
LGTM!!!!!!

PR #623, opened 2026-09-03
fix(upload): validate override filename

### LINKED ISSUE #622: Path traversal via the filename header bypasses the upload-directory sandbox
Affected versions: confirmed on commit `e18f15dee8bec8b1445d51cb978763bea1a2961d` (v0.18.0)

## Summary

rustypaste's upload handler accepts an optional custom `filename` HTTP header that overrides the stored file's name. Unlike the filename taken from the multipart `Content-Disposition` field (which is stripped down to a bare file name before use), the value of the `filename` header is used completely unvalidated. It is applied to the destination path with `PathBuf::set_file_name()` after the path has already passed the application's own directory-escape check, so a value containing `../` sequences lets an uploader write the paste's content to an arbitrary path on the filesystem, outside the configured `upload_path`.

## Details

`src/paste.rs`, `Paste::store_file()`:

```rust
let mut path =
    util::safe_path_join(self.type_.get_path(&config.server.upload_path)?, &file_name)?;
...
path.set_file_name(file_name);
if !no_extension {
    path.set_extension(extension);
}
if let Some(header_filename) = header_filename {
    file_name = header_filename;
    path.set_file_name(file_name);
}
...
let mut buffer = File::create(&path)?;
buffer.write_all(&self.data)?;
```

`util::safe_path_join` is the application's defense against traversal: it joins the base directory with a candidate name, lexically cleans the result, and rejects anything that no longer starts with the cleaned base. That check runs once, against the original file name (the multipart field name, already reduced to a bare file name via `PathBuf::file_name()` a few lines earlier in the same function). It is never run again.

The `header_filename` value, however, comes straight from the request's custom `filename` header with no processing at all (`src/header.rs`, `parse_header_filename`). When present, it overwrites `file_name` and is handed to `path.set_file_name()`, which does not validate or canonicalize its argument. If the argument contains its own `/`-separated components (including `..`), those components are appended verbatim to the path, after `safe_path_join`'s check has already passed on a different, unrelated value. The OS resolves the resulting `..` components at file-creation time, so the write lands wherever the traversal points, constrained only by the filesystem permissions of the process.

This affects every paste type that is written via `store_file`: `File`, `Oneshot`, `ProtectedFile`, and `RemoteFile` (which calls `store_file` internally after downloading). It does not affect `Url` or `OneshotUrl` pastes (`store_url`), because that function applies the header override to `file_name` before calling `safe_path_join`, so the join's check still sees the traversal value there and correctly rejects it. That correct ordering elsewhere in the same file shows the vulnerable ordering in `store_file` is an inconsistency, not the deliberate general design of the feature.

The custom `filename` header is honored regardless of whether `auth_tokens` is configured. When no `auth_tokens` are set (the shipped default `config.toml` ships this commented out), any unauthenticated client can exploit this. When `auth_tokens` is configured, any client holding a valid upload token can still exploit it, since the header is processed identically either way.

Suggested fix: validate the final path (including any `header_filename` override) with `safe_path_join`, or an equivalent canonicalize-and-prefix-check, as the last step before `File::create`, rather than only validating the pre-override name. The sibling `store_url` function already applies the header override before calling `safe_path_join`, showing the correct ordering.

## POC

(available upon request)

## Impact

Any party able to reach the upload endpoint (unauthenticated by default, or any holder of a valid upload token when `auth_tokens` is configured) can write attacker-controlled content to any path the rustypaste process can write to, limited only by OS file permissions. Depending on deployment, this can mean overwriting or creating files such as SSH `authorized_keys`, cron files, systemd units, or the application's own `config.toml`, turning an upload primitive into a path to full server compromise.

Severity: High, CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:H/A:N (7.5) in the default no-auth configuration; PR:L when `auth_tokens` is configured. CWE-22 Improper Limitation of a Pathname to a Restricted Directory ('Path Traversal').

carfeii:
## Summary

Fixes #622: the custom `filename` HTTP header overrides the upload's stored name via `path.set_file_name()` after `safe_path_join` has already validated a different, unrelated name (the multipart field's own filename). `set_file_name()` does not canonicalize its argument, so a header value containing `../` sequences escaped the configured `upload_path`.

## Fix

Re-validates the header override through `safe_path_join` against the configured upload root, the same pattern `store_url` already uses correctly for its own header-filename override (that function applies the override before calling `safe_path_join`, so it was never affected).

## Testing

Built the patched binary and validated in an isolated Docker network:
- A benign upload (no header) still succeeds and lands inside `upload_path` as expected.
- The same upload with `filename: ../sensitive/pwned_by_traversal.txt` now fails server-side with `sensitive/pwned_by_traversal.txt is outside of ./upload` instead of writing outside the upload directory.
- Confirmed on disk: the escape-target directory stays empty after the traversal attempt; the benign upload's file is the only one present.

### DIFF
--- a/src/paste.rs
+++ b/src/paste.rs
@@ -211,10 +211,8 @@ impl Paste {
         }
         if let Some(header_filename) = header_filename {
             file_name = header_filename;
-            // Re-validate through `safe_path_join` rather than `set_file_name`, which does not
-            // canonicalize its argument and would let a `../`-containing header value escape
-            // `upload_path` after the earlier join already checked a different, unrelated name.
-            path = util::safe_path_join(self.type_.get_path(&config.server.upload_path)?, &file_name)?;
+            path =
+                util::safe_path_join(self.type_.get_path(&config.server.upload_path)?, &file_name)?;
         }
         let file_name = path
             .file_name()
--- a/src/paste.rs
+++ b/src/paste.rs
@@ -211,7 +211,10 @@ impl Paste {
         }
         if let Some(header_filename) = header_filename {
             file_name = header_filename;
-            path.set_file_name(file_name);
+            // Re-validate through `safe_path_join` rather than `set_file_name`, which does not
+            // canonicalize its argument and would let a `../`-containing header value escape
+            // `upload_path` after the earlier join already checked a different, unrelated name.
+            path = util::safe_path_join(self.type_.get_path(&config.server.upload_path)?, &file_name)?;
         }
         let file_name = path
             .file_name()
--- a/src/util.rs
+++ b/src/util.rs
@@ -182,18 +182,25 @@ pub fn sha256_digest<R: Read>(input: R) -> Result<String, ActixError> {
 
 /// Joins the paths whilst ensuring the path doesn't drastically change.
 /// `base` is assumed to be a trusted value.
-pub fn safe_path_join<B: AsRef<Path>, P: AsRef<Path>>(base: B, part: P) -> IoResult<PathBuf> {
-    let new_path = base.as_ref().join(part).clean();
+pub fn safe_path_join<B: AsRef<Path>, P: AsRef<Path> + Copy>(
+    base: B,
+    part: P,
+) -> IoResult<PathBuf> {
+    let new_path = base.as_ref().join(&part).clean();
 
     let cleaned_base = base.as_ref().clean();
 
     if !new_path.starts_with(cleaned_base) {
+        error!(
+            "{} is outside of {}",
+            part.as_ref().display(),
+            base.as_ref().display()
+        );
         return Err(IoError::new(
             IoErrorKind::InvalidData,
             format!(
-                "{} is outside of {}",
-                new_path.display(),
-                base.as_ref().display()
+                "{} is outside of the base directory",
+                part.as_ref().display()
             ),
         ));
     }
--- a/src/util.rs
+++ b/src/util.rs
@@ -182,10 +182,7 @@ pub fn sha256_digest<R: Read>(input: R) -> Result<String, ActixError> {
 
 /// Joins the paths whilst ensuring the path doesn't drastically change.
 /// `base` is assumed to be a trusted value.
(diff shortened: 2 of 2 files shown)


src/paste.rs
```diff
@@ -211,7 +211,10 @@ impl Paste {
         }
         if let Some(header_filename) = header_filename {
             file_name = header_filename;
-            path.set_file_name(file_name);
+            // Re-validate through `safe_path_join` rather than `set_file_name`, which does not
+            // canonicalize its argument and would let a `../`-containing header value escape
+            // `upload_path` after the earlier join already checked a different, unrelated name.
```
tessus:
```suggestion
```

A comment is not necessary here. The function itself already includes a comment in `util.rs`.

tessus (changes requested):
Please remove the comment.

tessus:
Also, the formatting is off. There is a reason why there is a checklist in the pull request template. Not sure why you deleted the entire template.

tessus:
Ok, this is a genuine report, but the impact is rather limited. rustypaste always adds a timestamp to the filename on disk. This means you can't overwrite any system files.

No doubt this should be fixed though.

During testing I've noticed that the `safe_path_join` function emits a problematic error message. It shows the user the upload directory. I don't think a user should see the location of the files.
Furthermore the other info shows an already partly cleaned path.

e.g. let's say you use a header like `-H "filename: ../../test/filename"` you get an error message like:

```
../test/filename is outside of ./upload
```

3 things:

1. the error message should show the path as is: `../../test/filename`
2. the error message should read: `<path> is outside of the upload directory`
3. an error/info message can be shown in the rustypaste server log like `<path> is outside of <upload_directory>`

tessus (approved):
Now all is good. @orhun what do you think?

tessus:
@orhun I think we should merge this and create 0.18.1 asap.

orhun (approved):
ship it!

PR #626, opened 2026-09-07
chore(release): prepare for v0.18.1

tessus:
new version after fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -167,7 +167,7 @@ GitHub: <https://github.com/dbohdan/ferripaste>
 
 3. Rustypaste-GUI
 
-A script for a mininal GUI client for `rustypaste` (powered by zenity)
+A script for a minimal GUI client for `rustypaste` (powered by zenity)
 
 GitHub: <https://gist.github.com/rukh-debug/cc42900f86e39cacef6f7a6ba77ebf58>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.18.1] - 2026-09-07
+
+### Fixed
+- Validate override filename by @carfeii in [#623](https://github.com/orhun/rustypaste/pull/623)
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.18.1] - 2026-09-07
+
+### Fixed
+- Validate override filename by @carfeii in [#623](https://github.com/orhun/rustypaste/pull/623)
+
+### New Contributors
+- @carfeii made their first contribution in [#623](https://github.com/orhun/rustypaste/pull/623)
```
orhun, line 17: Let's mention that this is a crucial security fix and users should upgrade immediately.

We should also c/p this to the release page as well.

tessus: Sure, I can do that. Although I don't want to create too much panic, because the impact is limited. Yes, one can upload files to random dirs on the server. This is bad. Especially because an adversary can exhaust the filesystem's space. But no system files (or rp's config file as stated in the report) can be overwritten with this method since rustypaste adds a timestamp as an extension to an uploaded filename.

I'll think of something.

tessus: FYI: you can create a security advisory under `Security and quality` tab of the repo. I can't do that in your repo.

orhun: weird... do I need to enable something I wonder

tessus: I think some things are only available to the owner of a repo. I also never know which things one can make available to collaborators. But I believe for personal repos, not ones in an org, a few things are just available to the owner.

tessus: Ok, I looked into a bit. So collaborators have limited access. Maintainers can do more, but this role is only available for repos in an org. e.g. I can't even create a discussion in the category `Announcements`.

So here is what I am planning to do. I am going to create a discussion in `General` with the title `Security release 0.18.1` that details the issue.

In the CHANGELOG and the release page I'll add a Warning section that also references the explanation.

P.S.: You should be able to move the discussion to announcements (this is how it looks in my test repo):

image

tessus: Done, please change the category of #627 to `Announcements`.

orhun (approved):

PR #633, opened 2026-09-16
fix(paste): retry random URL generation on collision

### LINKED ISSUE #609: Random URL collision aborts upload (409) instead of retrying with a new name
## Description

When a generated random URL (petname or alphanumeric) collides with an existing filename, the upload is aborted with `409 Conflict` instead of automatically regenerating a new random name and retrying.

https://github.com/orhun/rustypaste/blob/d5b77406389f546c6d3edcedcf9e0ae4bf72c56a/src/paste.rs#L194-L225

```rust
if let Some(random_url) = &config.paste.random_url {
    if let Some(random_text) = random_url.generate() {
        // ...
        file_name = random_text;
    }
    // ...
}
// ...
let file_path = util::glob_match_file(path.clone())
    .map_err(|_| IoError::other(String::from("path is not valid")))?;
if file_path.is_file() && file_path.exists() {
    return Err(error::ErrorConflict("file already exists\n"));
}
```

`random_url.generate()` is called exactly once, and if the resulting name already exists on disk, the whole request fails rather than picking a different name. This is a real improvement over the old behavior in #312 (where a collision on an expiring paste silently pointed the returned URL at the wrong file), but a hard failure still pushes the burden of retrying onto the client — a script uploading via `curl`, or a CI job, has no reason to expect or handle a `409` here.

## Why this matters in practice

For the default `petname` config (`words = 2`), name generation always draws from `petname::Petnames::large()` (hardcoded, not configurable via `words`/`separator`):

https://github.com/orhun/rustypaste/blob/d5b77406389f546c6d3edcedcf9e0ae4bf72c56a/src/random.rs#L33-L40

That word list is 36,577 adjectives × 5,899 nouns ≈ **215.8M** possible 2-word combinations. Using the birthday-paradox approximation (`1.1774 × sqrt(N)`), a 50% chance of at least one collision is reached at only **~17,300 uploads** — not an edge case for anyone running a busy or long-lived instance, self-hosted CI artifact dumps, log shipping, etc.

## Suggested fix

Wrap the generate → check-exists step in a small bounded retry loop (e.g. retry N times, generating a fresh name each attempt, before finally giving up with `409`). This would make collisions transparently self-healing for the vast majority of cases without any client-side changes, while still surfacing an error in the unlikely case the retry budget is exhausted (e.g. word list exhaustion under `alphanumeric` with a very short `length`).

Happy to open a PR for this if a bounded-retry approach sounds reasonable — wanted to check in first since I'm not sure if there's a reason the current single-attempt behavior was chosen deliberately (e.g. to keep `store_file` free of loops/backoff).

spignelon:
Fixes #609

## Summary

`store_file` called `random_url.generate()` exactly once; a collision with an
existing filename rejected the upload with `409 Conflict` instead of trying a
different name. With the default `petname`/`words = 2` config (~215.8M possible
names), the birthday bound puts a 50% chance of at least one collision at only
~17,300 uploads — not an edge case for a busy or long-lived instance, and a
caller like `curl` in a CI job or cron job has no reason to expect or retry a
`409` here.

**Commit 1** wraps the generate → exists-check step in a bounded retry loop.
The budget is configurable via a new `random_url.retry_count` option (defaults
to `5`, clamped to a minimum of `1` so `retry_count = 0` can't be set which would make uploads with `random_url` impossible). A caller-supplied filename (via the
`filename` header) or a config without `random_url` is not retried, since
neither produces a new name to try on the next attempt.

```toml
# defaults to 5 if omitted
random_url = { type = "petname", words = 2, separator = "-", retry_count = 5 }
```

**Commit 2** fixes an adjacent TOCTOU race raised in the issue discussion: the
exists-check and the file write were two separate steps with nothing locking
them together, so two uploads landing on the same name at nearly the same
instant could both pass the check and the second would silently overwrite the
first. Switches `File::create` to `OpenOptions::create_new`, so losing that
race now surfaces as a clean `409` instead of a silent overwrite (and cleans up
the orphaned `.password` sidecar file for protected uploads). This commit is
separable if a narrower PR is preferred — happy to drop it if you'd rather keep
this PR scoped to the retry behavior.

## Testing

- `cargo test` — full suite passes; the only pre-existing failures are 3 tests
  that need outbound network access to `orhun.dev`/`raw.githubusercontent.com`
  (unrelated to this change, present on `master` too in this sandbox).
- `cargo clippy --tests -- -D warnings` and `cargo fmt -- --check` — clean.
- Added 4 new unit tests in `src/paste.rs` using a deterministic 62-name
  `alphanumeric` space (retry succeeds, retry budget exhausted, header filename
  is not retried, `suffix_mode` extension isn't compounded across retries).
- Manually verified end-to-end against a running server with
  `random_url = { type = "alphanumeric", length = 1, retry_count = 100 }` and
  61/62 possible names pre-occupied: upload succeeded and picked the one free
  name; a further upload correctly got a clean `409` once the name space was
  exhausted.
- Manually verified the atomic-write fix by firing two concurrent uploads at
  the same forced filename (`filename` header): one got `200`, the other a
  clean `409`, and the stored content was never a silent mix/overwrite.

## Out of scope

`Paste::store_url` has no existence check at all today, so a URL-shortener
entry can be silently overwritten on collision — a different bug from this
issue, worth a separate follow-up.


tessus (commented):
LGTM. Did you use AI for this PR?

spignelon:
Yep, used Claude Code to help implement and test this one.

tessus:
In this case, please add a `Co-authored-by: Claude <claude@anthropic.com>` or whatever the AI agent string is to the commit message(s). You can also squash the commits into one (we'd do that anyway upon merging). Ususally it is better to not force-push in PRs, but in this case it's ok, since we don't have 20 commits and a lot of messages going back and forth.

spignelon:
Done - squashed into one commit with the co-author trailer added, force-pushed.

tessus (approved):
Thanks, LGTM.

If Orhun approves as well, we can merge.

